PR #63, opened 2024-12-07
fix Folder Not Empty Message but Appears Empty

andersonbryan:
Folder appeared empty in Finder on MacOS, but got message **IDS_NOTIFY_FOLDER_MUST_BE_EMPTY**.  Troubleshooting the issue, I found the hidden _.DS_Store_ operating system file was returned using listFiles() causing to code to consider the folder not empty.  Added isEmptyFolder() which ignores existence of operating system hidden files like _.DS_Store (Mac), Thumbs.db (Windows)_


git-moss:
Thanks.

Issue #101, opened 2025-09-01
SF2 (SoundFont) files generated by ConvertWithMoss fail to convert with Roland SVZ Sample Converter


douglas-carmichael:
When I attempt to convert a SoundFont (SF2) created by ConvertWithMoss with [Roland's SVZ Sample Converter](https://www.roland.com/us/support/by_product/fantom_6_ex/updates_drivers/) I get a "Read Error!" message and the resulting SVZ file is unreadable on a Roland Fantom keyboard.

However, when I use [Polyphone](https://www.polyphone.io) to open the SF2 and then export it with "Export soundfonts.." without any changes, I am able to successfully convert the SF2 to an SVZ file and open it on a [Roland Fantom-06](https://www.roland.com/us/products/fantom-06/).

I've analyzed both SF2 files using sf2-test and found key structural differences:

### Missing Global Zones (Primary Issue)
**ConvertWithMoss output (fails conversion):**
- Preset has only 1 PBAG (missing global zone)
- Instrument has 22 IBAGS (no global zone at start)

**Polyphone re-export (works):**
- Preset has 2 PBAGs with proper global zone (0/0/global)
- Instrument has 23 IBAGS starting with global zone (0/0/global)

### Additional Differences
- **sm24 chunk**: CWM output includes sm24 chunk (7,412,852 bytes), Polyphone export has none
- **PBAG count mismatch**: CWM shows 2 pbag elements but preset only reports 1 PBAG

### Hypothesis
Roland's SVZ converter likely expects standard SoundFont structure with global zones for proper parsing. ConvertWithMoss appears to be generating technically valid but non-standard SF2 files that omit the global zones.

### Files Provided
- `SH Morning Dew-CWM_OUTPUT.sf2` - Direct ConvertWithMoss output (fails)
- `SH Morning Dew-CWM_FIXED.sf2` - Same file after Polyphone re-export (works)
- `*.txt` files - sf2-test analysis output for both files
- SoundFont files in attached ZIP

[SH Morning Dew SF2 Tests.zip](https://github.com/user-attachments/files/22070754/SH.Morning.Dew.SF2.Tests.zip)
[SH Morning Dew-CWM_FIXED.txt](https://github.com/user-attachments/files/22070752/SH.Morning.Dew-CWM_FIXED.txt)
[SH Morning Dew-CWM_OUTPUT.txt](https://github.com/user-attachments/files/22070753/SH.Morning.Dew-CWM_OUTPUT.txt)

[...]

PR #102, opened 2025-09-01
Fix SF2 global zone compatibility with Roland SVZ converter (issue #101)

douglas-carmichael:
- Add ensureGlobalZone() method to AbstractGroupedZones
- Implement createGlobalZone() in Sf2Preset and Sf2Instrument
- Call ensureGlobalZone() in createPresetDataChunks()
- Add null safety for global zones without samples

Fixes issue where ConvertWithMoss-generated SF2 files failed to convert with Roland's SVZ Sample Converter due to missing global zones.


git-moss:
Why did you remove the wiriting of the SM24 chunk? It is only written if there are 24-bit input files and by removing it it means that 24-bit samples will be broken.

git-moss:
If the SVZ tool cannot handle 24 bit files you cannot simply drop the 24 part. What would be necessary is an option to convert the samples to 16 bit. Please remove that from the pull request. I can look into the conversion.

douglas-carmichael:
Just added selectable 24-bit -> 16-bit downsampling for SF2 output.

git-moss:
Thank you!

PR #132, opened 2026-03-25
Support for Fairlight CMI 3 voices

PythonBlue:
Attempted to work with the codebase enough to implement support for Fairlight CMI Series III voices as a source format! Resources I don't currently have will likely be required to make it available as a destination format, but in the meantime, hopefully any retro sample enthusiasts with the appropriate files can try out.

Notes:

-tuning may be severely off when converted, and my experience is that that depends on the destination format, so that would need looking into as well. This version of the code should have correct tuning for SFZ conversion, though.

-filter cutoff and many other bells-and-whistles parameters of the time are currently unsupported. Silver lining is that the original sampler used no interpolation, thus the filter in it was normally intended to only control the imaging frequencies and aliasing anyway.

-CMI IIx files and earlier are unsupported despite their using the same VC extension.


PythonBlue:
I see a new commit has been made regarding my fork. Sorry I never got around to refining the code myself, but rest assured, will keep your formatting tips in mind for the future, especially seeing as I do have a Synclavier Regen I want to try to roll out support for as well.

git-moss:
Will try to have this in the next update.

PR #145, opened 2026-06-10
Fix Elektron Tonverk creator producing presets the device rejects

douglas-carmichael:
The Tonverk discarded converted presets and imported the WAV files as
loose samples (showing the first WAV's name as the instrument name) or
failed loading with a multi-sample error. Fixes verified against a
reference implementation tested on Tonverk hardware:

- Write the preset file with the correct '.elmulti' extension (was
  '.emulti') which the device and the detector require.
- Physically trim samples to the zone start/end instead of writing
  'trim-start'/'trim-end' which the device only accepts for
  single-file multi-samples and rejects otherwise.
- Create the preset data after re-sampling and trimming so trim and
  loop positions are written in output coordinates. Truncate normal
  loop points, keep the exact length of short single-cycle loops to
  prevent pitch drift and clamp loops into the sample boundaries.
- Never write a velocity layer with velocity 0.0 which makes the
  device reject the whole preset; substitute the factory default
  0.49411765.
- Fix the inverted tuning direction of 'key-center' (the detector
  already reads it as 'tuning = pitch - key-center').
- Name samples following the Elektron factory convention
  'Name-VVV-NNN-note[-rrN].wav' so the references in the preset file
  always match the written WAV file names (zone names previously went
  through createSafeFilename only for the files, not the references).
- Write 'keep-looping-on-release = true' for looped samples so loops
  sustain through the release phase (omitted means stop on release).
- Quote names containing single quotes correctly; the previous escape
  was a no-op and produced unparsable preset files.
- Only write a sample chunk when a loop exists and default the
  instrument/broadcast audio chunks to off: factory WAV files contain
  only 'fmt ', 'data' and 'smpl' chunks (in that order) and the
  device's WAV parser is strict.
- Format floats with up to 8 significant digits like the factory
  presets.

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 18.0.0
 
-* Added support for Elektron Tonverk emulti.
+* Added support for Elektron Tonverk elmulti.
 * Added support for Omnisphere 3.
 * Added support for reading Roland S-50, S-330, S-550, W-30.
 * Added support for reading Roland S-750, S-770, S-760, DJ-70, DJ-70 MkII, and SP-700.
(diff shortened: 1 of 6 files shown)


git-moss:
Many thanks, looks good. I guess you have a device and did test it (since I don't)?

douglas-carmichael:
Yes, I do own one.

PR #149, opened 2026-06-21
Add Renoise instrument (XRNI) format support (read + write)

douglas-carmichael:
## Summary

Adds read and write support for the **Renoise instrument format (`.xrni`)** — a ZIP archive containing an `Instrument.xml` description plus FLAC samples under `SampleData/`. Created files use document version 34 (Renoise 3.5); document versions 24–34 can be read.

### What is translated
- Key & velocity mapping, root note, tuning (transpose + fine-tune)
- Volume (linear gain) and panning
- Loops (off / forward / backward / ping-pong)
- Amplitude envelope (AHDSR volume modulation device)
- Per-sample sampler filter — type (LP/HP/BP/band-stop), cutoff and resonance, plus a cutoff envelope
- Pitch envelope
- Round-robins (the `Cycle`/`Random` keyzone overlapping mode)
- The instrument comment ↔ description metadata

The filter is written as the native sampler filter inside the modulation set, including the required `SampleMixerModulationDevice` (which backs the modulation "input" view) so the instruments open correctly in Renoise rather than crashing the editor.

### Also included
- **Fix:** FLAC or OGG samples stored inside a ZIP archive could fail to decompress with a `mark/reset not supported` error. `AudioFileUtils.decompressToWav` now wraps non-markable streams in a `BufferedInputStream`. This also benefits the existing discoDSP Bliss and DecentSampler library import.

### Implementation
- New package `format/renoise`: `RenoiseDetector`, `RenoiseCreator`, `RenoiseTag`, `RenoiseValueConverter`, `RenoiseFilterType`.
- Registered the detector and creator in `ConverterBackend`.
- Documented in `documentation/README-FORMATS.md`; `CHANGELOG.md` updated under 18.2.0.

### Known limitations (documented)
- 32-bit FLAC source samples cannot be transcoded (the bundled FLAC decoder handles ≤24-bit), so such samples are skipped with an error.
- The cutoff-frequency↔Hz curve is an approximation (Renoise's normalized-cutoff to frequency curve is internal); the filter type, resonance and envelopes are exact.

### Verification
- Reads the Renoise 3.5 factory instruments without errors.
- Round-trips with mapping, tuning, loops, panning and filter preserved; the generated `Instrument.xml` validates against the official `RenoiseInstrument34.xsd`.
- Confirmed to load in Renoise 3.5.4.


douglas-carmichael:
Added a second commit with an **opt-in loop cross-fade**, kept separate so it's easy to judge on its own.

**Default is unchanged and fully faithful** — loops are written exactly as they come from the source. Renoise has no loop cross-fade parameter, so *only* when the existing "loop cross-fade" processing option is enabled does the Renoise creator bake the cross-fade into the looped sample audio.

Motivation: some sources define loops with a built-in discontinuity (e.g. a SoundFont whose loop-offset generators land the loop on a non-matching point). That's reproduced faithfully by default, but it clicks in Renoise's sample-accurate looping; the option gives a way to get seamless loops without changing the faithful default. Left entirely optional for your call.

douglas-carmichael:
A couple of notes for merging:

**1. Complements #150.** Reading an XRNI decompresses its (stereo FLAC) samples, which trips a pre-existing bug that halved stereo compressed samples on decompress — fixed separately in #150. This PR stands on its own, but converting XRNIs to uncompressed destinations only yields full-length samples with #150 merged as well, so the two go well together.

**2. Small overlap with #150 in `AudioFileUtils.decompressToWav`.** Both PRs touch that method but on different lines — this PR adds the mark/reset stream wrap near the top of the method, while #150 fixes the frame-count line ~20 lines below. Each is clean against `main` individually; whichever lands second may need a trivial one-line rebase. Their CHANGELOG entries also sit under different version headings (`18.2.0` here, `18.1.2` in #150) — versioning is of course your call.

git-moss:
Many thanks! Can you point me to test files? Also for the Polyend format? PM me on KVR.

git-moss:
Thanks! I you can remove it again if you want.

PR #152, opened 2026-06-25
Add Synthstrom Deluge instrument format support (read + write)

douglas-carmichael:
Adds support for the Synthstrom Audible Deluge instrument format (the XML `sound`/`kit` files stored on the SD card), both reading and writing.

## Reading
- Reads **both** firmware variants — the official (element-based) form and the community (attribute-based) form — and both `sound` (synth) and `kit` (drum) files.
- Translates the sample mapping (key ranges), the original root note (from the *transpose*/*cents* offset, or the fixed root note of a drum), the sample playback start/end and loop points, the loop mode (one-shot or loop), the reversed flag, the low- and high-pass filter (with the various filter modes), the amplitude envelope and the velocity-to-volume modulation.

## Writing
- Writes a multi-sample as a `sound` patch with a single sample oscillator whose zones become a `sampleRanges` list.
- Uses the **element-based form of the official v4 firmware** (`firmwareVersion="4.1.0-alpha"`), so the output loads on both the official v4 firmware and the community firmware — no community-only features are used.
- Mirrors the Deluge SD-card layout: the patch is written to a `SYNTHS` sub-folder and its samples to `SAMPLES/<name>/`, with the `fileName` references written relative to the card root.

## Envelope &amp; loop handling
- The amplitude envelope attack/decay/release times are converted using the Deluge's internal rate tables rather than an exponential approximation, so the times round-trip faithfully. Attack uses the attack-rate table; decay and release share the release-rate table.
- The Deluge has no native loop cross-fade parameter, so — exactly like the Renoise creator — a loop cross-fade is baked into the looped sample audio. This happens automatically whenever a forward loop carries a cross-fade; the amount comes from the existing *Set fixed loop-crossfade* processing option or from the source. With none set, the loop is written exactly as-is.

Registered in `ConverterBackend`; documented in `README-FORMATS.md`; CHANGELOG updated. Verified by building and running EXS24 → Deluge conversions and loading the results on Deluge hardware running community firmware 1.2.1 — confirming that the v4 element-based output loads and plays on the community firmware as intended (it is designed to load on both the official v4 and community firmware).

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * New: Added support for the Polyend Tracker (PTI) instrument format (thanks to Douglas Carmichael).
 * New: Added support for the Renoise instrument (XRNI) format (thanks to Douglas Carmichael).
+* New: Added support for the Synthstrom Deluge instrument format.
 * New: Added support for the Downloadable Sound format (DLS) - read only.
 * New: Added several new tags for category detection.
 * FLAC/OGG
--- a/documentation/README-FORMATS.md
+++ b/documentation/README-FORMATS.md
@@ -65,6 +65,7 @@ The following multi-sample formats are supported:
 * [SFZ](#sfz)
 * [SoundFont 2](#soundfont-2)
 * [Spectrasonics Omnisphere 3](#spectrasonics-omnisphere-3)
+* [Synthstrom Deluge](#synthstrom-deluge)
 * [TAL Sampler](#tal-sampler)
 * [Waldorf Quantum MkI, MkII / Iridium / Iridium Core](#waldorf-quantum-mki-mkii--iridium--iridium-core)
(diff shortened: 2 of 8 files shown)


douglas-carmichael:
If you'd like real-world files to test the reader against, the official Deluge OLED v4.1.0 factory card contents — a full set of `sound` and `kit` XML presets with their samples, in the official element-based form — can be downloaded here:

https://synthstrom-audible-deluge.s3.us-east-2.amazonaws.com/Deluge+OLED+V4p1p0+factory+card+contents.zip

The detector also handles the community (attribute-based) variant, so patches saved on community firmware should read in as well.

git-moss:
Hi, Douglas, it seems you are really on a roll! :-) Many thanks!
In case you are planning more formats, it would be good to sync up, so we do not inadvertently start to work on the same formats. Might be easier to exchange email addresses, you can PM me on KVR.
Cheers!

PR #153, opened 2026-06-27
Add Elektron Tonverk preset (TVPST) format support for reading and writing

douglas-carmichael:
## Summary

Adds support for the Elektron Tonverk preset format (`*.tvpst`). In contrast to the existing elmulti/eldrum mapping files, a `.tvpst` is a full preset that also carries the synthesizer parameters. Both directions were exercised against the factory preset library, and presets written by ConvertWithMoss have been confirmed to load and play on a physical Tonverk.

### Reading

All three generator machines are read into the multi-sample model:

- **Multi** – a multi-sample mapped to key- and velocity-ranges.
- **One-Shot** – a single sample mapped across the keyboard (normalized sample start/end and loop points are scaled to frames).
- **Drum** – a kit of eight voices, each on its own key with its own settings.

The amplitude envelope (AHD or ADSR), the multi-mode filter together with its envelope (the bipolar env-depth maps to the cut-off modulator), the sample loops, gain and panning are converted. Synthesizer-specific parameters that have no equivalent in the multi-sample model (arpeggiator, effects, global LFOs, modulation matrix) are intentionally not converted. Samples referenced by absolute device paths (`/mnt/sdcard/...`) are resolved relative to the preset's location on disk.

### Writing

Writes a **Multi** or **Drum** machine preset, selectable in the creator settings (Multi-Sample / Drum Kit / Auto-from-source).

The output mirrors the device's SD-card layout so the result can be copied straight onto a Tonverk: the preset is a flat file in `User/Presets/<name>.tvpst`, and its samples live separately under `User/Multi-sampled Instruments/<name>/`, referenced by their absolute device path (`/mnt/sdcard/User/Multi-sampled Instruments/<name>/<sample>.wav`). Both sub-trees accumulate across presets when a whole library is converted. (A `.tvpst` is only listed by the device when it is a flat file in `User/Presets` with its samples referenced absolutely; the earlier folder-per-preset/relative-name layout showed nothing on the device.) Samples are written as physically trimmed WAVs, with a sample chunk only when the zone loops and no zero-velocity zones, matching the device's expectations.

The full `[parameters]` block is generated from a neutral factory template with the converted parameters filled in. The template was captured from a factory preset and has been fully neutralized — effects bypassed, overdrive and vibrato off, the filter wide open with a centered (no-sweep) filter envelope, and every per-machine LFO set to a centered depth (no modulation) — so a converted preset plays the source sample cleanly with only the amplitude envelope and any source filter, free of inherited coloration or modulation.

Envelope times and the filter cut-off are stored as normalized values driven by internal firmware curves. The envelope-time maxima were **calibrated against hardware resamples**: the firmware was found to follow the same `seconds = max · normalized³` cube law the writer uses, so the seconds mapping is close, not merely a guess. Times within the device's range round-trip to within a few percent; times beyond the device's ceiling clamp to it. A Tonverk→Tonverk round-trip remains loss-less (the inverse functions are exact), while a conversion to/from a unit-based format (Waldorf Quantum/Iridium, Synthstrom Deluge) is a close approximation.

### Naming

Since every Elektron format ConvertWithMoss currently handles is a Tonverk format, the existing `ElektronMulti*` classes (the elmulti/eldrum mapping) are renamed to `TonverkMulti*`, so the `format/elektron` family is named by device and leaves room for future Elektron devices. The mapping format's user-facing name is also unified to **`Elektron Tonverk Multisample`** for both reading and writing — it was previously `Elektron Multi` when read but `Elektron Tonverk` when written, the latter easily confused with the new `Elektron Tonverk Preset`. CLI prefixes are unchanged (`Elektron`/`Emulti` for the mapping, `Tonverk` for the preset), so existing scripts and settings are unaffected.

### Testing

Verified on a physical Elektron Tonverk and against the factory preset library:

- **On hardware:** presets written by ConvertWithMoss and copied to the SD card are listed, load and play on the device; the flat-file layout and absolute sample references were confirmed against how the device itself saves presets.
- **Envelope calibration:** device resamples of presets with known normalized values were used to measure the firmware's attack/release curves and set the seconds mapping. A held-note resample confirmed the neutralized template produces a clean, un-modulated tone (flat sustain, no warble).
- `.tvpst` → Waldorf QPAT and → Synthstrom Deluge — amplitude/filter envelopes and the filter carry through.
- `.tvpst` → `.tvpst` round-trip reproduces the amplitude and filter envelopes, the filter and the mapping (exact for the stored normalized values).
- All three machines, including per-voice drum envelopes and the Auto engine selection.


git-moss:
@douglas-carmichael Hi Douglas, In case you are planning more formats, it would be good to sync up, so we do not inadvertently start to work on the same formats. Might be easier to exchange email addresses, you can PM me on KVR.

One thing missing is the documentation of the format capabilities in: **documentation/SupportedFeaturesSampleFormats.ods** (edited with LibreOffice Calc) I already added two lines for the 2 new formats (but not yet for this one). Could you please fill the rows? Thanks!

PR #158, opened 2026-06-28
Waldorf Quantum/Iridium: write QPAT sample paths relative to the preset

douglas-carmichael:
### Problem

A QPAT preset created by ConvertWithMoss did not reliably find its samples on a Quantum/Iridium:

- Placed on the **SD card**, the device showed the **"Find Sample Map"** screen and the samples had to be located by hand (on a busy patch this also noticeably taxes the CPU).
- Using the device's own **Export Patches → With Samples** to back the preset up produced an invalid, **doubled** sample path (e.g. `3:2:samples/…`) and the backup failed.

### Cause

Each sample-map path was written as an **absolute** path with a leading drive number (`4:samples/…`, `4:` = USB). A drive number matches only one physical drive, so a preset on any other drive cannot be resolved automatically. And when the device exports the preset itself, it prepends the drive it is reading the preset from onto the already-absolute path, doubling the prefix (`3:` + `2:samples/…` → `3:2:samples/…`).

### Fix

Write the sample path **relative** to the preset (`samples/…`, no drive number). The device resolves a relative path against the folder the preset was loaded from, so:

- The samples load automatically regardless of which drive (SD / USB / internal) the preset sits on — no more "Find Sample Map".
- The device's Export → With Samples no longer doubles a prefix, so presets back up cleanly.

Confirmed on Iridium OS 4: relative-path presets both import and export/back-up correctly.

The detector already strips any leading drive number on read, so existing presets that still carry one continue to load.

---

*Note: this replaces the earlier version of this PR, which added a "Sample location" drive-number drop-down. Hardware testing showed the drive number itself was the problem — it broke the device's own backup/export — so a relative path is the simpler and more robust fix, and it removes the need for the user to know or pick the correct drive.*


douglas-carmichael:
**[Corrected]** The hardware result in this comment was a false positive — the samples were already in internal memory from an earlier test — and the `3:` fix here is wrong. See my updated comment below for the verified drive map (`2:` SD card, `3:` internal, `4:` USB) and the reworked fix.

douglas-carmichael:
**Correction to my earlier comment** — that first "clean import" was a false positive (the samples were already in the Iridium's internal memory from a prior test). I've since tested properly on my Iridium (OS 4.0.5): `3:` is **not** the fix, and the drive map in the original description was wrong. Verified on hardware: `2:` = SD card, `3:` = internal memory, `4:` = USB drive. The original `4:` was correct for USB and only ever broke SD-card users.

@git-moss — this has to be fixed, and not as a nice-to-have. **Every** QPAT written to an SD card today is broken on load: the device cannot resolve its own samples and drops you into a manual "Find Sample Map" hunt on **every single preset**. That alone is a UX failure — but the bigger issue is performance. Sample-map resolution runs on the same CPU that is playing voices, so on a busy patch that forced lookup competes with the audio engine and can make it choke. A sampler that stutters when you load a card-based patch is a hardware-performance bug, not a cosmetic one.

I have reworked this PR to fix it at the source: a **"Sample location"** option (SD card / USB drive / internal memory, default SD) that writes the correct drive prefix, plus a CLI parameter — all tested on hardware across the three settings. This should go into the next release. Tell me what you would like changed.

git-moss:
@douglas-carmichael I was in dicsussion with Waldorf about this and they even did not have an answer. One would expect that if a sound is imported, that the importer would adjust that.

douglas-carmichael:
Reworked this PR after further hardware testing on an Iridium (OS 4).

The previous version added a "Sample location" drive-number drop-down. It turned out the drive number itself is the root problem: with an absolute `<drive>:samples/…` path the device doubles the prefix when you use its own **Export → With Samples** (e.g. `3:2:samples/…`), so presets can't be backed up on the device.

Writing the sample path **relative** to the preset instead fixes both the original "Find Sample Map" issue *and* the backup issue, on any drive, with nothing for the user to get wrong — so I dropped the drop-down in favour of the relative path. Updated description above.

douglas-carmichael:
Agreed — and since the device's importer doesn't adjust the path (it actually *doubles* the drive prefix on its own Export → With Samples, e.g. `3:2:samples/…`), a relative path is the one form that needs no adjustment: it resolves against the preset's own folder, so it loads on any drive and round-trips through the device's own backup cleanly. That's what the reworked PR does.

git-moss:
That's great! So they finally improved that!

PR #169, opened 2026-07-07
SoundFont 2: add option to keep length-mismatched stereo-linked samples as mono

douglas-carmichael:
### Problem

Some SoundFonts — in particular commercial E-mu banks such as the "Orbit" bank — carry unreliable stereo metadata: samples are flagged as left/right and linked (`wSampleLink`) to partners they are not actually the stereo halves of. When such a bank is converted, `combineToStereo` tries to weld these "pairs" back together.

`compareSampleFormat` already declines the merge when the two samples differ in pitch or sample rate, but a **length** mismatch only logged a warning and then merged them anyway — so two unrelated sounds ended up hard-panned into the left and right channels of a single stereo sample (the shorter one truncating the longer). Converting one such bank also produced a wall of warnings.

### Change (updated per review)

Per @git-moss's feedback that some banks contain genuine stereo pairs whose left/right channels differ in length, the length-mismatch behaviour is now **opt-in and off by default**:

- New **"Keep mismatched stereo samples as mono"** option (`Sf2KeepMismatchedStereoAsMono`, off by default). When enabled, a length mismatch declines the stereo merge just like a pitch/sample-rate mismatch, so the two samples are kept as the separate mono samples they actually are. **Off by default a length-mismatched pair still combines**, so genuine different-length stereo pairs are unaffected.
- The pitch / sample-rate / length "left and right samples do not match" notices are now shown only when the existing **Log unused SF2 generators** option is enabled, so a normal conversion of a quirky bank stays clean while the detail is still available on request.

Since the stereo-combining happens while *reading* the SoundFont, the checkbox lives in the SF2 source options (next to "Log unused SF2 generators"); happy to move it if you'd prefer it elsewhere.

Rebased onto current `main`. The diff is limited to `Sf2Detector.java`, `Sf2DetectorUI.java`, `Strings.properties`, and the changelog / format docs.

### Verification

Tested against a commercial E-mu "Orbit" SoundFont, converting every preset to WAV:

- **Off (default):** the 3 length-mismatched pairs are each combined into a single stereo file (unchanged from before this PR).
- **On (`Sf2KeepMismatchedStereoAsMono=1`):** those 3 pairs are instead kept as their 6 separate full-length mono samples (+3 files); the channel that was previously truncated keeps its full length.

Pitch- and sample-rate-mismatched pairs remain mono in both cases (unchanged).

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -44,6 +44,8 @@
   * Fixed: Save formatting of ampersand character when writing.
 * SoundFont 2 (thanks to Douglas Carmichael)
   * Fixed: "Marker" presets that reference no samples (commercial SoundFonts often include one or two named after the vendor or copyright, e.g. "E-mu Systems 2007") were converted into empty instruments. Presets without any samples are now skipped.
+  * Fixed: The "left and right samples do not match" notices (differing pitch, sample rate or length) are now only shown when the "log unsupported attributes" option is enabled, so a normal conversion of a quirky bank is no longer flooded with warnings.
+  * New: Added a "Keep mismatched stereo samples as mono" source option (off by default). Some SoundFonts - notably commercial E-mu banks - carry unreliable stereo links that flag two unrelated mono samples as a stereo pair; if their left and right halves also differ in length they were welded into a single stereo sample. When the option is enabled a length mismatch keeps the two samples separate as mono (a pitch or sample-rate mismatch always does). It is off by default because some banks contain genuine stereo pairs whose channels differ slightly in length.
 * TX16W
   * Fixed: First check if the referenced absolute sample file path exists before searching all local folders.
 * Waldorf Quantum/Iridium (thanks to Douglas Carmichael)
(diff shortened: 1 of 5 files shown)


git-moss:
I have several examples where the left/right channels have different lengths. I suggest to add this as a destination option (which is off by default).

douglas-carmichael:
Thanks! Updated per your suggestion: this is now an opt-in **"Keep mismatched stereo samples as mono"** option, **off by default**, so banks with genuine different-length stereo pairs keep combining exactly as before — only when it's enabled does a length mismatch keep the two samples separate (matching how a pitch or sample-rate mismatch already behaves).

Since the stereo-combining happens while *reading* the SoundFont (`combineToStereo` in the detector), I put the checkbox in the SF2 source options next to "Log unused SF2 generators" rather than on the destination side, as the destination formats never see the split-mono pair — happy to move it if you'd prefer it elsewhere.

Verified on the E-mu "Orbit" bank: off (the default) combines the 3 length-mismatched pairs as before; on keeps them as their separate full-length mono samples. Rebased onto current `main`.

PR #176, opened 2026-07-07
User interface: rename the "Process" button to "Processing"

douglas-carmichael:
## Problem

The lower-toolbar button that opens the sample-processing options dialog is labelled **"Process"**. As a bare imperative verb it reads like an action that *starts* processing the samples — which is confusing for new users, especially since the actual run is started by the **"Convert"** button directly above it, and this button only opens a configuration dialog.

## Fix

Rename it to **"Processing"** — a single word that:

- **is not an imperative** — "Process" is the command form; "Processing" is a noun, so it reads as a label for a screen, not an order;
- **matches the dialog it opens** — that window is titled "Processing" (`IDS_PROCESSING_DIALOG`), so button and title now agree;
- **parallels its neighbour "Settings"** — both are lower-toolbar buttons that open a configuration dialog and both carry the same gear icon;
- **keeps the layout unchanged** — same footprint as "Settings"/"Analyse".

Also adds the `Alt+P` mnemonic that its sibling buttons already have, and rewords the tooltip from "Configure processing of samples during conversion." to "Open the options for processing samples during conversion." to reinforce that it opens a dialog.

Resource-only change (`Strings.properties`); the button wiring in `MainFrame` is untouched (the `"Process"` argument there is the `Process.png` icon filename, not the label). CHANGELOG updated.


git-moss:
I changed this intentionally from Processing to Process since the longer name broke the layout on some platform/screen resolutions and I was unable to fix this. Until you checked this on all OSs and some resolutions I don't want to go through the hassle again.

PR #177, opened 2026-07-09
OGG: fix frame count reported as -1 and decoding losing the final block

douglas-carmichael:
### Problem

Two related defects for OGG sources (found converting the Vulture Culture "Legendary Synths" SFZ libraries, whose samples are OGG, to Waldorf QPAT):

1. **Frame count / bit resolution reported as -1.** Vorbis stores neither in its header, so `AudioSystem.getAudioFileFormat().getFrameLength()` returns `NOT_SPECIFIED` and `DefaultAudioMetadata` passed the -1 through. Every destination format that works with the sample length then wrote corrupt values. E.g. the Waldorf Quantum/Iridium creator normalizes loop points to [0..1] fractions of the frame count — dividing by -1 wrote them as large negative numbers (`-191785.00000000` instead of `0.59562964`), so the device could not resolve the sample maps and every converted preset showed the "Locate Samples" screen.

2. **Decoding lost the final block.** The Java Sound wrapper around the JOrbis decoder stops draining it at the last whole block and ignores the end-of-stream length trim given by the granule position of the last Ogg page. Every decoded OGG came out slightly short (451 frames ≈ 10ms on the test files), so sample loops ending at (or near) the end of the file lost their loop end and conversions could never be sample-exact.

### Change

- `OggFileSampleData.createAudioMetadata`: when the provider reports -1, the frame count is read from the **granule position of the last Ogg page** (for Vorbis this is exactly the total number of PCM sample frames). The tail scan requires the candidate page to end exactly at the end of the file, which rules out false matches inside compressed payload data; if no valid page is found it falls back to fully decoding the file. A bit resolution of -1 becomes the 16 bit the decoder produces.
- New `OggVorbisDecoder`: drives the same JOrbis engine that is already bundled in `java-vorbis-support` (no new dependency), fully drains the synthesis state after each block and so emits **all** sample frames — the output length matches the granule position. `OggFileSampleData.writeSample` uses it, keeping the Java Sound path as a fallback for files the direct decoder cannot parse (nothing has been written to the output stream in that case). Since every conversion path funnels through `writeSample` (including `rewriteFile` → `convertToWav`), all destination formats benefit.
- `WaldorfQpatCreator`: normalized sample start/end and loop positions are clamped to the [0..1] range the device expects — source formats may reference positions slightly beyond the length of the audio data (e.g. loop points authored for the original sample but shipped with a lossy-compressed file).

### Verification

- **SFZ+OGG → QPAT** (187 presets, 2 commercial libraries): previously every sample map contained negative loop values; now all map fields validate in range and loop points land sample-exact — e.g. an authored loop `191785..321972` comes out as fractions that multiply back to exactly those frames.
- **Decoded audio:** for the frames the old path produced, the new decoder's output is **bit-identical** (same engine, same float→16-bit conversion); only the missing tail is added. The recovered tail correlates 0.998 with the same region of the library's original WAV files (the OGGs' pre-encode source). Output frame count equals the granule position exactly.
- **Non-OGG regression:** converting a 234-preset SF2 (E-mu Orbit) to QPAT on this branch and on current `main` produces **byte-identical** output for every preset and sample file.
- The granule tail-scan and the [0..1] clamp are exercised by all 3841 OGG samples of the two libraries (loop points beyond the decoded length occur in practice there).


git-moss:
Might be worth to submit to the JOrbis project as well.

PR #179, opened 2026-07-11
Add support for the Roland MV-8000/MV-8800 patch format (MV0)

douglas-carmichael:
## What

Adds read **and** write support for Roland MV-8000 / MV-8800 patches (`*.mv0`) — the pad-based Production Studio workstations (2003/2006). A patch bundles all parameters and the sample audio in a single file: up to 96 partials assigned to the 96 pads via a note table, each partial layering up to 4 SMT sample slots.

## How the format was obtained

Roland never documented the format. It was reverse-engineered in two stages:

1. **Statistically from the 103 factory patches** (field positions located by corpus-wide correlation: velocity splits, stereo pairs, root keys, loop points etc.).
2. **Confirmed against the firmware**: the MV-8000 OS 3.54 / MV-8800 OS 1.01 updater images contain per-object parameter descriptor tables (byte/bit offset, bit width, default, min, max for every packed parameter). These matched all empirical findings, corrected three fields (pan/coarse/key-follow) and provided the whole TVF/TVA/LFO block. The MV-8800 table is byte-identical to the MV-8000's, so both machines share the format.

The complete field-level documentation is included as `documentation/design/MV8000_FORMAT.md` (happy to move it elsewhere if you prefer a different location).

## Supported

- Patch name + category (the format stores the XV-5080 category list; mapped to the tag detector categories)
- Note table → key ranges; per-SMT velocity range + crossfade
- Pitch key-follow (drum kits/menu patches are fixed-pitch, instruments chromatic; fractional key-follow mapped to key tracking)
- Play modes: loop / one-shot; loop start/end
- SMT level, pan, coarse/fine tune
- TVA amplitude envelope (T1-T4/L1-L3) and velocity on/off; TVF filter: type (LPF/BPF/HPF), cutoff, resonance, envelope + depth
- 16-bit/44.1 kHz embedded sample data; stereo samples (stored as hard-panned L/R mono pairs with a `0x7F L/R` name suffix) are combined to stereo zones on reading and split back on writing; identical samples mapped to several key ranges are stored once

Limitations (documented in README-FORMATS): the hardware curve of the envelope times is unknown (the sound engine runs on a separate CPU whose program is not part of the OS updater), so times are approximated with the S-7xx curve; MV-8800 `*.mvf` files are MFX effect presets (e.g. the Analog Bass Synth), not patches, and are not convertible.

## Verification

- All 103 factory patches convert to SFZ and other formats without errors (correct velocity splits, loops, fixed-pitch kits, stereo merging, envelopes/filters spot-checked against musically obvious cases: pads have slow attacks, the piano decays to zero sustain, the filter-sweep basses carry full envelope depth).
- MV0 → MV0 round-trip: a single-sample factory patch reproduces at the identical file size with bit-exact PCM; the factory Grand Piano round-trips its stereo pairs, sample names, loop points, envelope and slot parameters exactly.
- `documentation/SupportedFeaturesSampleFormats.ods` has the new rows filled in.

CHANGELOG and README-FORMATS are updated.


git-moss:
I noticed there is an unused MV8000Patch.CATEGORIES. Can this be removed or should it be used for category detection?

PR #180, opened 2026-07-12
Add support for the Roland ZEN-Core sound format (SVZ)

douglas-carmichael:
Adds a detector and creator for the Roland ZEN-Core `.svz` sound format, shared across the FANTOM-0, FANTOM / FANTOM EX, GAIA-2, Juno-X, Jupiter-X/Xm and the ZENOLOGY plug-in. The undocumented container and tone format was reverse-engineered from device exports and the FANTOM / FANTOM-0 firmware, with the tone-parameter offsets validated against the factory tone database. Format notes are in `documentation/design/ZENCORE_FORMAT.md`.

**Write** — an importable `.svz`, loaded on the device via *UTILITY -> IMPORT*:
- a single tone for one multi-sample, or a multi-tone bank sharing one sample pool for several;
- per tone the source's filter (type / cutoff / resonance) and TVA amplitude envelope;
- per sample the name, key range, root key, loop and level;
- user samples written at the device-native 48 kHz / 16-bit, mono duplicated to stereo.

**Read** — the user samples and multisample key map of a `.svz` are read back.

**Loop clicks / sample preparation** — the ZEN-Core voice engine has no loop cross-fade or de-click of its own and clicks whenever playback crosses a non-zero sample value (the device's own sampler behaves identically). For click-free hardware playback the loop start and end are snapped to the nearest zero-crossing and the post-loop tail is tapered to silence before writing.

**Target Device option** — the `.svz` header's 5-character model tag is selectable: `KY019` (the shared ZEN-Core tag of the FANTOM / FANTOM-0 / FANTOM EX / Juno-X / Jupiter-X/Xm; the default, and the only device-confirmed case here), `MI085` (GAIA-2) or `RC001` (ZENOLOGY plug-in). The GAIA-2 and ZENOLOGY tags are derived from the device firmware, not yet confirmed on hardware.

`README-FORMATS.md`, `CHANGELOG.md` and `SupportedFeaturesSampleFormats.ods` are updated.


git-moss:
Nice one! Thank you!

PR #199, opened 2026-07-18
ACCESSIBILITY: dialogs open with a bright white flash (photosensitivity risk) and are unstyled in dark mode

douglas-carmichael:
Flagging this one for attention: sudden full-window white flashes against a dark UI are a genuine accessibility risk for photo-sensitive users (epilepsy, migraine, many neurodivergent users) - the kind of event WCAG's flash criteria exist for. Since it fires on every open of the Settings and Processing dialogs, dark-mode users are exposed to it constantly.

Two causes, both in `MainFrame.setDarkMode`:

1. The dark stylesheet was only added to the scene of the **main window**, so the Settings and Processing dialogs are not styled at all in dark mode - they render in the light theme against the dark main window.
2. Every window kept the default **white scene fill**. JavaFX renders the fill until the first pulse paints the content, so each dialog (and the main window at startup) briefly shows a pure white frame - very visible in dark mode.

Fix: `setDarkMode` now applies/removes the stylesheet on the dialog scenes as well (they are reachable via `AbstractDialog.getWindow ().getScene ()` and exist from construction) and pre-fills all three scenes with the theme background color (#373E43 dark / #ECECEC light). The first rendered frame then already has the theme color, so the white flash cannot be rendered anymore in either mode.

Note: message boxes created inside uitools (`Functions.message`) would need the same treatment in uitools itself to be fully covered - happy to provide that as well if wanted.

Verified so far: compiles, and the application starts cleanly with the new `setDarkMode` path running at startup (the dialog scenes exist at that point, no exceptions). Visual confirmation on macOS is being done by the reporter and I will follow up on this PR with the result.

Stacked on #195 (contains its commit) only so the CHANGELOG 19.0.1 section does not conflict - the code change itself is independent of it.

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -11,6 +11,8 @@
 * TAL Sampler (thanks to Douglas Carmichael)
   * Fixed: The sample "reverse" flag was never read as enabled: it is stored numerically (0/1) like all other TAL flags but was parsed as a true/false text boolean.
   * Fixed: Disabled groups were only skipped when written as enabled="0" but not as enabled="false". Presets that switch between several kits via a drop-down in their UI (each kit is a group and only one is enabled) were converted with all kits stacked on the same keys and playing at once.
+* User Interface (thanks to Douglas Carmichael)
+  * Fixed: The Settings and Processing dialogs opened with a bright white frame and, on macOS, repainted whenever the main window was clicked - both a risk for photo-sensitive (e.g. epileptic) users. The dialog scenes are now pre-filled with the theme background color (no white frame), the dialogs are independent, non-modal, always-on-top windows so clicking the main window no longer repaints them, they carry the base and dark-mode stylesheets directly (previously inherited only by owned dialogs), and the main action buttons are disabled while a dialog is open.
 
 ## 19.0.0
 
--- a/src/main/java/de/mossgrabers/convertwithmoss/ui/MainFrame.java
+++ b/src/main/java/de/mossgrabers/convertwithmoss/ui/MainFrame.java
@@ -27,6 +27,7 @@ import de.mossgrabers.convertwithmoss.core.INotifier;
 import de.mossgrabers.convertwithmoss.core.creator.ICreator;
 import de.mossgrabers.convertwithmoss.core.detector.IDetector;
 import de.mossgrabers.convertwithmoss.core.settings.ICoreTaskSettings;
+import de.mossgrabers.tools.ui.AbstractDialog;
 import de.mossgrabers.tools.ui.AbstractFrame;
 import de.mossgrabers.tools.ui.EndApplicationException;
 import de.mossgrabers.tools.ui.Functions;
@@ -40,6 +41,7 @@ import de.mossgrabers.tools.ui.panel.ButtonPanel;
 import javafx.animation.PauseTransition;
 import javafx.application.Platform;
 import javafx.beans.binding.Bindings;
+import javafx.beans.binding.BooleanBinding;
 import javafx.collections.FXCollections;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
@@ -49,6 +51,7 @@ import javafx.geometry.Orientation;
 import javafx.geometry.Pos;
 import javafx.scene.Node;
 import javafx.scene.Parent;
+import javafx.scene.Scene;
 import javafx.scene.control.Button;
 import javafx.scene.control.ComboBox;
 import javafx.scene.control.ListView;
@@ -66,6 +69,7 @@ import javafx.scene.layout.HBox;
 import javafx.scene.layout.Pane;
 import javafx.scene.layout.Priority;
 import javafx.scene.layout.StackPane;
+import javafx.scene.paint.Color;
 import javafx.stage.Stage;
 import javafx.util.Duration;
 
@@ -166,8 +170,22 @@ public class MainFrame extends AbstractFrame implements INotifier
     {
         super.initialise (stage, baseTitleOptional, true, true, true);
 
-        this.settingsDialog = new SettingsDialog (this.getStage ());
-        this.processingDialog = new ProcessingDialog (this.getStage ());
+        // The dialogs are created without an owner and non-modal. On macOS a dialog that is
+        // owned by (or modally blocking) the main window is repainted by the window manager
+        // whenever the main window is clicked. An independent window is not, so clicking the
+        // main window is an ordinary focus change with nothing to repaint. They are kept on top
+        // so they cannot get lost behind the main window, and the main action buttons are
+        // disabled while a dialog is open (see below) to keep the one-at-a-time behavior.
+        this.settingsDialog = new SettingsDialog (null);
+        this.processingDialog = new ProcessingDialog (null);
+        keepOnTop (this.settingsDialog);
+        keepOnTop (this.processingDialog);
+
+        // An owned dialog inherits the base spacing/padding styles from its owner window; since
+        // these dialogs are unowned they must load the base stylesheet themselves, otherwise
+        // their layout falls back to the tighter default spacing
+        this.addBaseStyles (this.settingsDialog);
+        this.addBaseStyles (this.processingDialog);
 
         // The main button panel
         final ButtonPanel upperButtonPanel = new ButtonPanel (Orientation.VERTICAL);
@@ -182,6 +200,16 @@ public class MainFrame extends AbstractFrame implements INotifier
         this.processingButton.setOnAction (_ -> this.openProcessing ());
         this.settingsButton = setupButton (lowerButtonPanel, "Settings", "@IDS_MAIN_SETTINGS", "@IDS_MAIN_SETTINGS_TOOLTIP");
         this.settingsButton.setOnAction (_ -> this.openSettings ());
+
+        // The dialogs are modal, but macOS does not always block a click on the owner window
+        // reliably, so the main action buttons are additionally disabled while either dialog is
+        // open - a disabled button emits no action, so neither dialog can be opened twice or
+        // re-triggered while present (which previously caused a repaint flash)
+        final BooleanBinding anyDialogOpen = this.settingsDialog.showingProperty ().or (this.processingDialog.showingProperty ());
+        this.convertButton.disableProperty ().bind (anyDialogOpen);
+        this.analyseButton.disableProperty ().bind (anyDialogOpen);
+        this.processingButton.disableProperty ().bind (anyDialogOpen);
+        this.settingsButton.disableProperty ().bind (anyDialogOpen);
 
         // Source pane
         //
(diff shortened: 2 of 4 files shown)


douglas-carmichael:
Added a second commit which closes the remaining gap discussed in the description: a `Window.getWindows()` list change listener now applies the current theme (stylesheet + scene fill) to every stage the moment it is shown. This also covers the message boxes created inside uitools (`Functions.message` etc.), which the application could not style before - so no window renders an un-themed white frame anymore, including error/info popups. Popup windows (tooltips, combo box lists) are excluded because a solid fill would break their transparency.

The proper library-level fix for uitools is here: git-moss/UiTools#1 (dialogs inherit the owner window styling). If that lands in a uitools release, the listener and the explicit dialog styling here become redundant but stay harmless - take whichever combination you prefer.

douglas-carmichael:
Third commit after feedback from testing on macOS: with the theme fill alone the opening of the Settings/Processing dialog was still visible as step-by-step painting (filled empty frame first, then the controls). Windows are now kept at opacity 0 while they open and revealed 80 ms later fully rendered - handled centrally in the window list listener, so it covers the two dialogs, the uitools message boxes and the application start. 80 ms is below the perception threshold for a delay but enough for layout, styling and positioning to complete.

douglas-carmichael:
Fourth commit, after a report that the Processing dialog still flashed on re-open while the Settings dialog was fine. I instrumented the window timeline (opacity/geometry listeners plus post-layout pulses) and found a race: the window list listener fires only after the window is already mapped, so the opacity-0 arrives up to ~10 ms after the first presented frame. Whether that raced frame is visible depends on the window - the Processing dialog re-shows a 765 px high stale backing and flashes, the small Settings dialog slips through unnoticed.

The main window and the two re-used dialogs now set their opacity to zero whenever they are hidden (and before their first show), so the window is already fully transparent when it is mapped - the race is structurally gone, verified with the same instrumentation (on re-open no opacity transition is observed after mapping anymore). The window list listener remains as the best-effort path for windows created inside uitools. Also guarded the reveal against a window which is closed again within the reveal delay, so it cannot start its next show visible.

douglas-carmichael:
Fifth commit - the reporter clarified that the remaining flash happens when the Process button is clicked while the Processing dialog is already open. Instrumentation confirmed the mechanism: on macOS the click is not reliably blocked by the application-modal dialog, so the handler ran again, reset every control of the live dialog to the stored settings (a visible repaint of the whole dialog - and it also silently discarded any edits made so far) and then showAndWait threw "IllegalStateException: Stage already visible" on the FX thread. The Settings dialog has the same code path but few controls, so its reset slipped by unnoticed. Both handlers now just focus the already open dialog and return.

douglas-carmichael:
Sixth commit. The click-while-open guard from the previous commit still called `requestFocus()` on the already open dialog, and that re-order/repaint was itself an intermittent flash. Root fix: the Process and Settings buttons are now disabled via a binding to their dialog`s `showingProperty()` while the dialog is open, so the button cannot be triggered again at all (a disabled button emits no ACTION). Verified with instrumentation - firing the button 5x while the dialog is open no longer enters the handler (previously it entered, hit the guard and called requestFocus). The early-return guard is kept as a safety net for an event queued before the disable took effect, now without requestFocus. This is the last of the flash paths I can find; a modal owner-button on macOS is best neutralised by disabling it rather than trying to swallow the re-entrant call.

douglas-carmichael:
Seventh commit, and the one that targets the actual mechanism. With the button disabled the flash still happened on a real click, which ruled out any app-side handler: it is macOS re-rendering the modal dialog when its blocked owner window is clicked (it scales with dialog size - the large Processing dialog flashes, the small Settings dialog does not). The Settings and Processing dialogs are now non-modal, so the window manager has no blocked window to flash. To preserve the one-at-a-time behavior the four main action buttons (Convert, Analyse, Process, Settings) are disabled while either dialog is open. Verified the non-modal showAndWait path and the button enable/disable transitions with instrumentation; the reporter is confirming the visual result.

douglas-carmichael:
Eighth commit - and a course correction. A tester reported the opacity-reveal made things worse: clicking anywhere on the main window could trigger a visible opacity transition that reads as a flash. The opacity gate (opacity 0 on hide + 80ms reveal via the window-list listener) is removed entirely. What remains is the low-risk, high-value core: dialogs inherit the dark stylesheet, every dialog scene is pre-filled with the theme background color so the first rendered frame is never white, the dialogs are non-modal (no macOS modal-attention flash), and the main action buttons are disabled while a dialog is open. Net effect: no white frame and no modal flash, without any opacity animation. The uitools-level fix (git-moss/UiTools#1) still covers the library message boxes.

douglas-carmichael:
Ninth commit, reverting my own non-modal change. Testing showed the non-modal dialogs introduced a worse flash: on macOS, clicking the owner window while a non-modal *owned* dialog is open re-renders (flashes) the dialog. Modal dialogs block clicks on the owner, so that does not happen. The dialogs are modal again (the original behavior). The re-entrancy repaint that motivated non-modal is already handled independently by disabling the main action buttons while a dialog is open, so that stays. Net remaining change vs stock: dialogs inherit the dark stylesheet, scene fill is pre-set to the theme color (no white first frame), and the action buttons are disabled while a dialog is open. No modality change, no opacity animation.

git-moss:
Good luck with that, I tried to fix this 2 weeks ago...

douglas-carmichael:
Update: the reporter has confirmed on macOS that the repaint is gone with the final approach - the Settings and Processing dialogs are independent, unowned, non-modal windows (kept always-on-top, with the main action buttons disabled while a dialog is open). Clicking the main window is now an ordinary focus change with no owned/modal child to repaint. I will push the final code (currently the branch still has the intermediate modal revert) once a minor dialog-sizing follow-up is settled.

douglas-carmichael:
Final code pushed and confirmed by the reporter on macOS: the repaint is gone AND the dialogs render with the correct (roomy) spacing. The last piece was that owned dialogs inherit the application base stylesheet (DefaultStyles.css) from the owner window; making them unowned dropped that, so they now load DefaultStyles.css (and the dark-mode stylesheet) directly. The commit history shows the iteration to get here (opacity reveal and modal/non-modal attempts, all superseded) - the net diff is: theme scene-fill, independent non-modal always-on-top dialogs, base+dark stylesheets applied to the dialog scenes, and the action buttons disabled while a dialog is open. Happy to squash if you prefer a single commit.

douglas-carmichael:
Rebased onto current main (after #195/#196/#198/#200 merged) - now a single clean commit and MERGEABLE. The fix is confirmed working on macOS by the reporter: no white frame, no repaint on main-window clicks, and the dialogs render with the correct spacing (they load the base stylesheet directly since, being unowned now, they no longer inherit it from the owner).

git-moss:
Latest source is now checked in. I tested on Windows 11, Linux Ubuntu, and latest MacOS but could not notice any of the described issues. Dark/light mode switches always apply to all dialogs as well. The only issues I am aware of is the "flashing" on the main window on startup. Can you explain how to reproduce that?

douglas-carmichael:
Sure — and the short answer is that you cannot reproduce it on the checked-in source because you already fixed the structural part of it yourself.

**The flash as users experienced it (reproducible on every release up to 18.1.1):** enable dark mode, quit, relaunch — it must be persisted; the runtime toggle never exercises this path, since it restyles windows that are already rendered. Up to 18.1.1 the uitools `AbstractFrame.initialise ()` ended with `stage.show ()`, so the stage became visible before `MainFrame` built its content and before `loadConfiguration ()` added `Darkmode.css`. Every dark-mode start therefore rendered: empty white window → light-styled UI → dark restyle, on all platforms. Your "Improved user interface" commit from 2026-06-25 (`this.show ()` at the end of `MainFrame.initialise`, with the matching uitools change) removed exactly that sequence, and 19.0.0 — released the day this PR was opened — was the first build containing it. The startup-flash reports come from 18.1.1-and-older builds; to see it today, run 18.1.1 in dark mode.

**What remains on current source is much smaller:** the gap between the OS mapping the window and JavaFX presenting the first frame shows the native window backing, and the main scene keeps `Color.TRANSPARENT` as fill (uitools `initScene ()`), so nothing covers that gap FX-side either. On Windows the backing is white; on macOS it follows the *system* appearance. Repro conditions: application dark mode + OS in **light** appearance + cold start (first launch after login; `-Dprism.order=sw` or parallel CPU load stretch it). On a warm machine it is a single frame — record the screen at 60 fps and step frames. Testing dark mode on a dark desktop maps a dark backing, so there is nothing bright to flash — I suspect that is why it does not show for you. The Mac where this was reported runs the system in light appearance with the application in dark mode, which is the worst case.

The scene-fill part of this PR narrows that remaining window on the FX side (first-layout gap shows theme color instead of white — the same mechanism that made the dialogs flash, since a `Dialog` scene has a white default fill). The native pre-first-frame backing itself is not reachable from JavaFX, so that last frame can only disappear when OS appearance and app theme match.

douglas-carmichael:
Tested on macOS. Both symptoms are back with 85de63c — and I could isolate which change causes which.

**Setup:** macOS 26.5.2 (Apple Silicon), system appearance **Light**, application dark mode **on** (persisted). Built 85de63c locally and packaged it with `jpackage:jpackage@mac` — bundled JDK 26.0.1, JavaFX 26.0.2, uitools 2.0.6. Run as an installed `.app`, not from an IDE.

**Test 1 — 85de63c as-is:** Settings and Processing both open with the white flash again, and the repaint when clicking the main window is back as well.

**Test 2 — 85de63c plus the scene fill only.** Your dialogs left exactly as you have them, owned and `APPLICATION_MODAL`; the only change is re-adding the fill in `setDarkMode`:

```java
final Color fill = Color.web (isSelected ? "#373E43" : "#ECECEC");
this.scene.setFill (fill);
this.settingsDialog.getWindow ().getScene ().setFill (fill);
this.processingDialog.getWindow ().getScene ().setFill (fill);
```

**→ White flash gone. Repaint-on-click still present.**

So these are two independent problems with two independent fixes, and the rollback removed both:

- **White flash when a dialog opens** = the scene fill. Without it the fill stays `Color.TRANSPARENT` (uitools `AbstractFrame.initScene`), so the first frame the window presents is white. This has nothing to do with ownership or modality — Test 2 is the proof, since it is your current code with nothing changed but the fill.
- **Repaint when clicking the main window** = owned + `APPLICATION_MODAL` on macOS. The window manager repaints a modal dialog whose blocked owner is clicked; an independent window is not, so the click is an ordinary focus change with nothing to repaint.

**Suggestion**, in the shape you already started with the button binding:

1. Restore the fill unconditionally. It is platform-neutral, it cannot cause the repaint issue you saw (Test 2 is that exact configuration), and it should fix the dialog flash on Linux too.
2. Gate only the ownership/modality on `OperatingSystem.isMacOS()` — the class you added in this same commit. macOS: unowned, non-modal, always-on-top, action buttons disabled while a dialog is open. Windows and Linux: owned + `APPLICATION_MODAL`, exactly as now.

Regarding your `// TODO test` on the `isMacOS()`-gated button binding: it only does anything in the non-modal configuration. While the dialogs are modal the buttons cannot be clicked anyway, so right now it is an inert safety net; it becomes load-bearing again if macOS goes non-modal.

And agreed on Windows — the native window painted before JavaFX takes over cannot be fixed from FX. That is a separate thing from the FX-side white first frame, which is what the fill addresses; I am not claiming the fill does anything about the native one.

Happy to open this as a PR against current main if you want it in that shape, and I can turn around any macOS test you want quickly.

PR #200, opened 2026-07-18
Synthstrom Deluge: add an option to write a drum kit

douglas-carmichael:
The Deluge creator only ever wrote a synth (*sound*) preset, even though the detector already reads both *sound* and *kit* files. This adds the missing direction: a drum-kit writer.

A new **Output Type** creator option (a dropdown with *Synth (Sound)* / *Drum Kit*; CLI `DelugeOutputType=sound|kit`, default `sound`) selects what is written. Modelled on the existing Tonverk "Output Engine" option. A kit turns every zone into its own drum (a `<sound>` inside the kit`s `<soundSources>`), laid out low to high; the preset is written to a `KITS` folder with its samples alongside in `SAMPLES/<name>/`, mirroring the SD-card layout the sound path already uses.

The type is chosen **explicitly rather than auto-detected**: a one-sample-per-note layout is not necessarily a kit - a per-note-sampled synth bass looks identical - so guessing would misclassify those.

I cross-checked the kit structure against the Deluge firmware source (`Kit::writeToFile` / `SoundDrum::writeToFileAsInstrument`) and a factory kit: the `<kit>` root carries the same `firmwareVersion` / `earliestCompatibleFirmware` attributes as a sound, followed by a kit-level `<defaultParams>` block and one `<sound>` (with `<name>`) per drum. Only v4-compatible, element-based structure is used, matching the sound writer.

A kit holds at most one drum per note (36 up to 127), so at most 92 drums are written; a larger source is truncated with a logged warning instead of silently.

Verified by converting a 40-drum DecentSampler kit to a Deluge kit and reading it back through the Deluge detector as 40 drums, and by confirming an over-size source is capped to 92 with the warning. Documented in README-FORMATS.md, CHANGELOG entry added.


douglas-carmichael:
Added a second commit: velocity layers and round-robins are now consolidated by note. A Deluge kit drum is a single sample, so a source with several zones on the same note previously produced several duplicate drums. The kit path now reuses the synth path`s `getMappedZones`, keeping one zone per note (the loudest velocity layer). Verified with a synthetic kit of two notes each carrying a soft and a loud layer -> two drums, each the loud layer; the 40-note SSS043 kit is unchanged at 40 drums.

PR #201, opened 2026-07-18
Synthstrom Deluge: kit consolidation + natural-pitch fix

douglas-carmichael:
Stacked on #200 (drum-kit output) - please merge that first; this PR then reduces to a single commit.

A prototype for compacting large drum kits into a Deluge-friendly layout. Opt-in (CLI `DelugeConsolidateKit=1`, off by default), only meaningful when Output Type = Kit.

What it does:
* Reduces the kit to one drum per recognized type (several kicks/snares/toms collapse to the first).
* Orders the drums by drum role - kick on the lowest row - following the factory **TR-808** layout, so kick/snare/tom/hats sit on the lowest rows and a beat can be programmed without scrolling through rows.
* Unrecognized drums are all kept and appended after the recognized ones (nothing is silently dropped; a summary is logged).

The classifier recognizes drum types from the name, handling full names, numbered variants (`kic2`, `sna3`) and the four-character truncations the Deluge factory kits store (`cras`, `cowb`, `tamb`, `shak`, `bloc`, `caba`, `tria`, `spla`, `chin`, `timl`/`timh`, ...). I surveyed the drum-name vocabulary across all 43 factory kits on a card to build it, so it is not 808-specific.

Validated by converting a 40-drum DecentSampler kit (-> 15 drums) and by round-tripping the factory R-50 kit through the detector (19 -> 16 drums, correctly classified and role-ordered).

This is explicitly a **prototype for feedback** - the drum-role order and the type vocabulary are easy to tune. Documented in README-FORMATS.md, CHANGELOG entry added.

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Synthstrom Deluge (thanks to Douglas Carmichael)
   * New: Added an Output Type creator option (Synth/Kit, CLI DelugeOutputType) to write a drum kit instead of a synth (sound) preset. A kit writes one drum per note, consolidating velocity layers and round-robins to the loudest layer (a Deluge drum is a single sample). The type is chosen explicitly because a one-sample-per-note layout is not necessarily a kit (e.g. a per-note synth bass).
+  * New: Added a "Consolidate kit" option (CLI DelugeConsolidateKit) which reduces a drum kit to one drum per type (kick, snare, hi-hat, ...) ordered by drum role following the factory TR-808 layout (kick on the lowest row), so a beat can be programmed without switching rows. The consolidated drums are labelled by their role for a clean read-out on the device.
+  * Fixed: Kit drums were pitched by their keyboard mapping note, so a drum mapped to e.g. note 35 played 25 semitones too low (audible on toms). Kit drums now play at their natural pitch (transpose 0), like the factory kits; only an explicit detune is kept.
 * DecentSampler (thanks to Douglas Carmichael)
   * New: Added a source option "Create one multi-sample per group": creates a separate multi-sample for each group (disabled groups included), e.g. for presets which contain several alternative kits as groups and switch between them via their user interface.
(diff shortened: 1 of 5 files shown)


douglas-carmichael:
Rebased onto current main now that the drum-kit feature (#200) is merged - this is a single clean commit on top of main and now shows MERGEABLE. It contains two things:

1. **Consolidate kit** (the original opt-in feature, `DelugeConsolidateKit=1`): one drum per recognized type, ordered by drum role (kick on the lowest row, TR-808 layout), role-labelled for a clean OLED read-out, sample files keep their names.
2. **Bug fix folded in**: kit drums were pitched by their keyboard mapping note, so a drum mapped to e.g. note 35 played 25 semitones too low - clearly audible on toms. Kit drums now play at natural pitch (transpose 0), like the factory kits (verified: all factory 808 drums use transpose 0). Only an explicit detune is kept. This touches the same writeSound/writeZone path as the consolidation, which is why it is in this PR rather than a separate one; happy to split it out if you prefer.

douglas-carmichael:
Tested the kit-name shortener against real freely-available libraries (downloaded the FreePats "Synthesizer Percussion" SFZ set, plus a survey of the karoryfer/DrumGizmo/Salamander free kits and the Deluge factory names). Two results:

- The consolidation classifier generalised well to an independent library: FreePats (names like ClosedHiHat, MidTom, Claves) consolidated 19 -> 9 correctly, role-ordered and role-labelled.
- The shortener revealed one real gap - it left "SynthesizerPercussion-20220718" unchanged (camelCase + date suffix). Added a safe fix: a trailing date (6-8 digits) or version ("v2") suffix is now stripped, while short model numbers (TR-808, R-50, SDS-5) are preserved (they are 1-4 digits), and a name that is itself only "Kit NN" is no longer double-numbered. FreePats now shortens to "SynthesizerPercussion"; SSS043 ("007 Full Kit 2") and the factory model-number names are unaffected. I deliberately did not add CamelCase-splitting (it lengthens names, defeating the purpose) or hard truncation (lossy, collision-prone).

PR #204, opened 2026-07-19
Roland ZEN-Core: read the compact SMPd sample header (SVZ version 02 01)

douglas-carmichael:
Fixes a read error reported for the ZEN-Core SVZ format: mono/stereo detection was wrong and the sample-length calculation crashed the in-memory sample copy.

### Cause

The reader assumed the FANTOM/KY019 `SMPd` layout for every file: a 460-byte header with the channel count in the byte at `+0x08`. Some SVZ files (SVZ file version `02 01`, e.g. third-party sample pools) use a different, compact `SMPd` encoding:

| | FANTOM / KY019 | Compact (version `02 01`) |
|---|---|---|
| Header size | 460 (`0x1CC`) | **158 (`0x9E`)** |
| Channel count | byte at `+0x08` | **u16 at `+0x0A`** |
| Byte at `+0x08` | the channel count | **`0x32`** (unrelated) |

So the channel count was read from the wrong field (a constant `0x32` → the mono/stereo mistake), and the PCM was located 302 bytes off. Because the two header sizes differ by 302 (`2 mod 4`), the wrong header left the PCM two bytes short of a whole stereo frame, so the WAV buffer ended up smaller than the PCM and the copy threw.

### Fix

`ZenCoreSvz.resolveSmpdLayout` detects the layout per chunk: it picks the header whose channel-count field (byte `+0x08` for the 460 header, u16 `+0x0A` for the 158 header) is 1 or 2 **and** whose PCM region past the header divides evenly into frames. The 302-byte size gap makes exactly one header valid for a stereo chunk, so no version flag is needed and an unrecognized chunk is left without audio rather than crashing. `decodePcm` now copies only whole frames as a further guard. The declared play length is deliberately not used to size the PCM, since device exports store a few frames more or fewer than `f04 / 2`.

The third `SMPd` encoding is documented in `documentation/design/ZENCORE_FORMAT.md`.

### Verification

Read every sample through the `InMemorySampleData` → WAV path (the reported crash site): 164/164 samples write cleanly across the reported file (32 stereo loops) plus every device export in the reverse-engineering corpus; reverting the fix reproduces the crash on the reported file. No change to the write path — the FANTOM/KY019 exports still read as before (460-byte header, channels at `+0x08`).

### DIFF
--- a/documentation/design/ZENCORE_FORMAT.md
+++ b/documentation/design/ZENCORE_FORMAT.md
@@ -325,6 +325,33 @@ Culture's *SOURCE*): a `"SMPd"` tag with a u16 header size `0x20` at +4 and u8 v
 is byte-identical to the device's. CWM decodes such packs and re-emits device-native raw-PCM
 `SMPd`.
 
+**Compact `SMPd`** (file version `02 01`; seen in third-party sample-pool exports, e.g. a pool
+tagged `Nemly`): follows the same `f04 = 2 × end` law at +4, the same bits at `0x009`, rate at
+`0x00C` and name at `0x010` as the device-native encoding, but uses a shorter **158-byte**
+(`0x9E`) header with a shorter preview, and stores the **channel count as a u16 at `0x00A`** — the
+byte at `0x008` (channels in the device encoding) holds an unrelated `0x32`. Interleaved 16-bit
+LITTLE-endian PCM follows at `0x9E`. CWM reads it; it is not written.
+
+```
+0x000   4  "SMPd"
+0x004   4  u32 f04 = 2 × end
+0x008   1  0x32  (NOT the channel count in this encoding)
+0x009   1  bits (16)
+0x00A   2  u16 channels (2)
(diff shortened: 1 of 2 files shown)


douglas-carmichael:
On the second point from the report — the `USP_CHANNELS` field looking like "a very high value", suggesting the write side is also wrong:

That one is a red herring. The `USPa` record stores the channel count as a **single byte at +0x2C**, and it is correct (`2`) — it matches the device's own `uspa.bin` template and every FANTOM export (`STEREO_TEST`, `EXPORT_SAMPLE`, `DC_FINALTEST`, …), as well as the reported file. The high value appears only if the field is read as a wider integer: +0x2E/+0x2F hold `32 32` (the sample-format region), so a u16/u32 read starting at +0x2C spills into them (e.g. a u32 reads `0x32320002`).

So the write path is correct, and this PR deliberately leaves it unchanged — the fix is entirely on the read side.

PR #212, opened 2026-07-20
Carry seven more instrument parameters through the model (one-shot, choke group, amplitude key-tracking, random play logic, ...)

douglas-carmichael:
Seven parameters which many formats can express but the model could not carry, so they were silently dropped on every conversion. Each is one model commit plus one roll-out commit, so any of them can be taken or dropped on its own.

Selected by counting how many formats can actually supply or consume the value - the aim was portability, not covering a single device.

| Parameter | Level | Formats | Why it was needed |
|---|---|---|---|
| One-shot playback | zone | 22 | Every format collapsed it into "has no loop", or guessed it from the envelope |
| Exclusive ('choke') group | zone | 11 | `DlsRegion.getKeyGroup ()` was fully documented with zero callers; `RenoiseTag.MUTE_GROUP_INDEX` was referenced by nothing; SoundFont 2 generator 57 had no constant at all |
| Amplitude keyboard-tracking | zone | 8 | The counterpart of the existing filter cutoff keyboard-tracking |
| `PlayLogic.RANDOM` | zone | 6 | Random selection was collapsed into ALWAYS, i.e. all layers sounding at once |
| Envelope time key/velocity scaling | envelope | 8 | Distinct from the existing slopes, which describe the curvature of a segment rather than its duration |
| Per-instrument voice settings | instrument | 10 | DecentSampler synthesised a monophonic preset from a user-interface option because the model had no value |
| Group volume/panning/tuning | group | 6 | Detectors flattened these into every zone and creators then had nothing to write back |

**Every default reproduces the previous behaviour exactly**, so no creator output changes until a format opts in. Verified: a Deluge to Waldorf conversion of 100 presets is byte-identical to `main`, samples included.

### Round-trip proof

The same SFZ through `main` and through this branch:

| | `loop_mode` | `amp_keytrack` | `polyphony` |
|---|---|---|---|
| main | `no_loop` | dropped | dropped |
| this branch | `one_shot` | `0.500` + `amp_keycenter=60` | `4` |

### Notes on the design

* **The amplitude key-tracking unit is fixed to 1 dB per key** relative to the root key. Unlike pitch and filter key-tracking there is no natural 1:1 reference, so without a fixed unit each format would pick its own anchor and the value would not survive a conversion. Formats whose native law is unknown approximate their range linearly and say so at the conversion site.
* **A format which cannot express a random selection falls back to round-robin, never to ALWAYS** - playing every layer simultaneously is much further from the intent than cycling through them. `IGroup.hasRoundRobin ()` and `isFullRoundRobin ()` therefore accept RANDOM as well, while the two sequence-based methods keep requiring ROUND_ROBIN, since a randomly selected zone has no sequence position.
* **The group values are only recorded in addition to the existing flattening into the zones**, which is unchanged. Only the Waldorf creator consumes them, because the per-oscillator volume and panning have no per-zone equivalent in that position; it subtracts the hoisted offset from the sample map so the net value is identical.
* **Two read sites deliberately require explicit evidence**, because a wrongly set one-shot means note-off is ignored and notes never stop: Akai MPC treats a missing attribute as a one-shot (that fall-back still drives the loop handling but is no longer stored), and Elektron Tonverk writes loop mode 'Off' for every sample which simply has no loop, so it is not read back as one.

### Waldorf Quantum/Iridium - a behaviour change worth calling out

This is the one place where output changes for existing sources, and it fixes a real loss:

* The oscillator volume and panning were always written as 0 dB / Center, so a Quantum/Iridium round trip lost them.
* The detector *overwrote* the sample map's gain and panning with the oscillator values instead of combining them - the actual cause of the loss.
* A preset without an oscillator panning panned every zone half right, because the converted [-1..1] value was initialised with a value of the raw [0..1] range.

Measured on 235 real presets exported from an Iridium, comparing the net per-zone gain (converted to SFZ so the comparison is on values rather than on the internal split):

| | net gain preserved |
|---|---|
| main | 37 / 235 |
| this branch | **235 / 235** |

### Also included

`DefaultFilter.equals` compared the cutoff velocity modulator twice instead of comparing the cutoff envelope modulator. Two filters differing only in their cutoff envelope therefore counted as equal, so `getGlobalFilter ()` combined zones which are not identical, and a filter with a cutoff envelope but no velocity modulation threw a `NullPointerException`.

* `SupportedFeaturesSampleFormats.ods` documents the new parameters in six new columns: volume/panning/tuning in the Group section, a Voices column (polyphony, monophonic, legato), one-shot and exclusive group in the Sample section, and Key Tracking plus Envelope Time Tracking in the Amplitude section - filled separately for the read and the write direction of every format with the native parameter names. The random play logic needed no new column, the affected formats already show their play-logic support.

### Still outstanding

* `IModulator.getCurve` is part of the model but no format reads or writes it yet - the curve types of the Roland S-7xx and MV-8000 are unlabelled graphs in the manuals, and the Yamaha YSFC, Tonverk and disting EX enumerations are undocumented. It is deliberately left unwired rather than guessed, and is not mentioned in the change log because it changes no output.


git-moss:
@douglas-carmichael There are some conflicts now after accepting your other pull requests. Could you please update them?
I won't touch SupportedFeaturesSampleFormats.ods any time soon, so please update that as well.

douglas-carmichael:
@git-moss Done:

* Merged main and resolved the conflicts. All your changes are kept: the DecentSampler sequence length stays the highest round-robin position (now under the extended random/round-robin flags), the restructured MV-8000 slot loop received the two new setters, and the change-log entries are combined under the shared headings. The merged branch compiles and the effective diff against main is unchanged apart from the resolutions.
* `SupportedFeaturesSampleFormats.ods` now documents the new parameters in six new columns: "volume / panning / tuning" in the Group section, a "Voices" column (polyphony, monophonic, legato), "one-shot" and "exclusive group" in the Sample section, and "Key Tracking" plus "Envelope Time Tracking (key / velocity)" in the Amplitude section. Each is filled separately for the read and the write direction with the native parameter names; granularity limits are amber ("only global", "per group"). The random play logic needed no own column - the affected formats already show their play-logic support.

One observation while filling the matrix: the sheet has no rows at all for the Synclavier Regen (so its one-shot and group offsets from this PR are not shown) and none yet for the new Kurzweil format. I can supply complete rows for both in a follow-up if you want.

douglas-carmichael:
The two missing formats now have complete rows in the matrix as well (derived from their detector/creator code):

* **Synclavier Regen** (between SFZ and Synthstrom Deluge): timbre title/comment/#tag metadata, partials as groups with their pan/volume/tune offsets, patch-entry mapping (key range, root, start/end and loop fractions, one-shot loop bits), the per-partial VEnv amplitude envelope, the velocity crossfade windows and the global note filter incl. PitchTrack. Library write = the 64-timbre library folder; the loop cross-fade is only a flag in the format (bit 3), so it is amber on the write side.
* **Kurzweil K2000/K2500/K2600** (between Korg KMP/KSF and Kontakt): programs/keymaps/velocity levels, root key, entry tuning plus layer transpose, VolumeAdjust and the forward loop. Write-side ambers where the format quantizes (8 dynamic levels) or the value is baked into the audio (play start/end, the audio cut at the loop end); envelopes and filters are red since the converter does not read or write them.

PR #214, opened 2026-07-22
Restore the dialog scene fill on all platforms, apply the dialog window changes only on macOS

douglas-carmichael:
Follow-up to #199 and your rollback in 85de63c, with the macOS testing you asked for.

The two symptoms have **two independent causes**, which is why rolling back the modality change also brought back the white frame. I isolated them with packaged builds on macOS 26.5.2 (Apple Silicon, system appearance Light, application dark mode on, bundled JDK 26.0.1, JavaFX 26.0.2, uitools 2.0.6):

| Build | White frame on dialog open | Repaint when clicking the main window |
| --- | --- | --- |
| 85de63c as-is | **yes** | **yes** |
| 85de63c + scene fill only (dialogs left owned and `APPLICATION_MODAL`) | no | **yes** |

So:

- **The white frame is the scene fill**, nothing else. Without it the fill stays `Color.TRANSPARENT` (uitools `AbstractFrame.initScene`) and the first frame a window presents is white. It is platform independent and has nothing to do with ownership or modality — the second row is your current code with only the fill added.
- **The repaint is the owned, modally blocking dialog.** The window manager repaints a modal dialog whose blocked owner is clicked; an independent window is not, so the click becomes an ordinary focus change with nothing to repaint.

This PR therefore splits them along the line you already started with the button binding:

1. The scene fill is restored **unconditionally**, for the main window and both dialogs. This cannot cause the repaint problem you saw — the isolation build above is exactly that configuration — and it should fix the dialog flash on Linux as well.
2. Only the **ownership and modality** are gated on `OperatingSystem.isMacOS()`, the class you added in 85de63c. macOS gets unowned, non-modal, always-on-top dialogs with the main action buttons disabled while one is open. **Windows and Linux keep owned, `APPLICATION_MODAL` dialogs exactly as they are now**, so the systems where non-modal misbehaved for you are untouched.

`SettingsDialog` and `ProcessingDialog` take the modality from the caller instead of hard-coding it; that is the only change to those two files.

Regarding your `// TODO test` on the `isMacOS()`-gated button binding: it only does anything in the non-modal configuration, since modal dialogs block the buttons anyway. It is kept inside the macOS branch where it is load-bearing.

I also updated the CHANGELOG entry from #199, which still described the dialogs as independent and non-modal on all platforms.

And to be clear about Windows: I am not claiming anything about the native window painted before JavaFX takes over. You are right that it cannot be fixed from FX. The fill addresses the separate FX-side white first frame, which is what the dialogs show.

Happy to adjust the shape, and I can turn around any further macOS test quickly.


douglas-carmichael:
Confirmed on macOS 26.5.2 (Apple Silicon) with a packaged build of this branch (b3e1c07 via `jpackage:jpackage@mac`, system appearance Light and application dark mode on, which is the worst case here): no white frame when the Settings or Processing dialog opens, and no repaint when the main window is clicked while a dialog is open. Both symptoms are gone.

git-moss:
I fixed this now by implementing them as pseudo dialogs. There is now no flashing at all and it does not require OS specific code. Tested on all 3 OSs.

PR #220, opened 2026-07-24
Added support for the E-mu Emulator IV bank format (E4B) - not yet tested on hardware

douglas-carmichael:
This adds a detector and creator for the banks of the E-mu Emulator IV series (Emulator 4, E4X, E4XT, E4K, e-Synth, e-6400 and the other EOS samplers): `*.e4b`.

## Reading
Every preset of a bank becomes one multi-sample and every voice becomes a group. Mapped: names, key and velocity ranges, root keys, tuning (transpose + coarse + fine, plus the per-zone fine tune and the sample pitch offset), volume, panning, sample-level forward loops, the amplitude envelope (attack/hold/decay/sustain/release plus its velocity modulation from the cord table) and the filter (type/poles, cutoff, resonance, key tracking and the filter envelope with its depth). The EOS effect filter types (phasers, flangers, vocal formants, EQ morphs) have no model equivalent; such voices are converted without a filter. Banks written by older EOS versions (no EMSt chunk, oversized FORM size) are read as well.

Banks are also read directly from CD-ROM and hard disk images of the EOS samplers (`.iso`, `.img`, `.hda`), either from the E4B source format or through the ISO/IMG one. The EOS samplers read no standard filesystem from CD-ROM at all, so nearly all Emulator IV content in circulation exists as images of E-mu's proprietary filesystem. The geometry is taken from the superblock, so the CD and the hard disk variant are both read; EIII banks on discs of the older samplers are skipped.

### Naming
The presets of the commercial EOS libraries are named after the articulation or the variation they provide (*Dark Tremolo*, *Long Release*, *Natural Range*) while the instrument they actually play is only given by the name of their bank, so the converted multi-samples were neither identifiable nor unique - a name like *Natural Range* exists in most banks and those presets overwrote each other. The bank name is now prepended unless the preset name already starts with it (*Greek Bazouki - Dark Tremolo*, but *Greek Bazouki XS* unchanged), which is switchable, and each bank gets a folder of its own when the folder structure is created. Independently of that the bank is passed on as the description, which is where the formats with a field of their own for the bank take it from - the Waldorf Quantum/Iridium for example shows it next to the preset name, which suits a device with little room for a long name.

Two smaller read fixes came with it: every bank of those libraries ends with an unusable placeholder preset (`Emu Systems 2000`) whose zone references a sample that is not in the bank, which was dropped correctly but logged an error for each of its zones - a zone without a sample (index 0) is skipped silently now and a lost zone is only reported for a preset which does contribute other zones. And the name fields of the disk filesystem are padded with zero bytes rather than spaces, which showed up as `Greek Bazouki???`.

## Writing
Each multi-sample becomes one preset; a library collects all sources into a single bank (up to 1000 presets and 1000 samples, the format limits). Every zone is written as its own voice, which keeps all per-zone settings; the panning, which has no voice level counterpart, is written into the zone entry. Samples are stored as 16-bit mono PCM at their original sample rate (stereo sources are mixed down with a note in the log) and are de-duplicated by content. Rates above 48 kHz, the EOS maximum, are down-sampled with the zone and loop positions moved along. Since the sampler always plays a sample back at 44.1 kHz, the pitch offset which transposes a sample stored at any other rate back into tune is written with it. The root note is appended to the sample name (e.g. `_C3`) as the hardware browsers show it.

The destination can additionally write the bank as a ready-to-use CD-ROM image for SCSI CD-ROM emulators such as the ZuluSCSI, as a checkbox next to the bank output. Only EOS 4.7+ reads FAT hard disks, so for anyone on an older EOS version an image with the proprietary filesystem is the only way to get banks onto the sampler, and even on 4.7 the CD-ROM route is the most common workflow.

## Format knowledge and attribution
The layout was reverse-engineered by the mpc2emu project (GPL, hardware-saved E4XT banks, commercial EOS CD-ROMs, the emu3bm sample struct and Phil's E4 notes); this implementation is an independent one based on that documentation. A distilled spec with the byte layouts and the hardware-calibrated envelope rate law lives in `documentation/design/E4B_FORMAT.md`.

## Validation
- Banks written by the creator pass the mpc2emu reference parser plus structural checks (EMU FORM-size convention, trailing EMSt master setup, TOC offsets); the EMSt/E4Ma blocks are byte-identical to hardware-captured defaults.
- Three hardware-created third-party banks (Ian Wilson's free EOS banks, 198 presets / 1980 zones / 317 sample chunks) read completely: every zone resolves and all 165 unique sample PCMs are byte-identical to the reference parser's output.
- A 66-preset bank round-tripped through detector and creator re-parses with identical zone mapping, envelopes and filters.
- The three per-zone offsets are new ground - mpc2emu writes zeros for those bytes - and were recovered from the E-mu Producer Series Volume 7 CD-ROMs (2336 presets, 76057 zones), with the evidence recorded in `documentation/design/E4B_FORMAT.md`: the fine tune at `12:14` only ever holds `0x00`/`0xFF` in its high byte, exactly matching the sign of the low byte, which identifies it as one signed big-endian value; measuring the recorded pitch of the samples against the root key of their zone shows it to be a correction of that deviation, and applying it as 1/64 semitones drops the average pitch error of a bank from 12.3 to 7.7 cents (as cents it would only reach 8.7). The volume at `15` has the small signed range of a level trim and is used by the mixed presets which balance their zones, the panning at `16` spans the full -64..+63 and its users are exactly the stereo spreads (*Wide*, *Percussion Section*, *Congas/Bongos*) while the plain single instrument presets leave it at 0.
- A bank round-tripped through detector and creator keeps the panning and the volume byte-identical and is pitch-exact in the tuning; the pitch offset of a sample moves into the voice fine tune there, since the model has no sample level tuning of its own.
- Two defects which the mpc2emu project found on real hardware after this PR was opened are fixed here as well: the sample pitch offset, without which a sample not stored at 44.1 kHz plays transposed, and a CD-ROM image directory entry which could describe a bank one block short so that the sampler refuses to load it. The image change leaves every already-correct entry byte-identical; see the discussion below for the verification of both.

I do not own an Emulator IV, so written banks have **not been loaded on real hardware yet** - same caveat as the MC-707 support. Happy to adjust anything.


douglas-carmichael:
Added the feature matrix rows for the format to SupportedFeaturesSampleFormats.ods (Read/Write pair between discoDSP Bliss and Elektron Tonverk). While filling it in I also tightened the creator: written samples are now limited to 48 kHz (the maximum EOS playback rate) - higher-rate sources are down-sampled with the zone/loop positions moved accordingly.

douglas-carmichael:
Two additions since the initial version, both aimed at making the format actually usable on real hardware - especially for owners of **older units**:

**Reading EOS CD-ROM and hard disk images** (*.iso*, *.img*, *.hda*). The EOS samplers read no standard filesystem from CD-ROM at all, so nearly all Emulator IV content in circulation exists as images of E-mu's proprietary disk filesystem (dumps of the commercial CD-ROM libraries, images for SCSI emulators like the ZuluSCSI). Banks are now read directly from such images, following the existing house pattern for sampler disk images: the endings on the E4B source format itself plus an EMU3 entry in the ISO/IMG format identifier which dispatches to the E4B detector (like the Ensoniq and Roland handling there). The filesystem geometry is taken from the superblock, so the CD and the hard disk variant are both read; EIII banks on discs of the older samplers are skipped.

**Writing a CD-ROM image** as a destination option, following the DecentSampler 'Create Bundle' checkbox pattern. The rationale: only EOS 4.7+ can read FAT hard disks, so for anyone running an older EOS version an image with the proprietary filesystem is the *only* way to get banks onto the sampler - and even on 4.7 the CD-ROM route via ZuluSCSI is the most common workflow. The written images are **byte-identical** to those of the mpc2emu reference builder for the same bank content, whose images are verified to load on a real E4XT, so the filesystem layer inherits that hardware validation (the bank content itself remains untested on hardware, as noted above).

I had briefly opened the image reading as a stacked PR (#222) but folded it in here so the format can be reviewed as one piece.

lentferj:
Check my latest update, especially 
https://github.com/lentferj/mpc2emu/commit/e6b85b8d417a1f01398a1397379694fe764bfde1
https://github.com/lentferj/mpc2emu/commit/2d5f8326cf6734c6667ecff59232c18172817761
https://github.com/lentferj/mpc2emu/commit/c40b7eec846a0071dd9a1faf305873e3dc9f65e7

douglas-carmichael:
Thank you for the pointer - both of those were present here too, and both are now fixed on the branch.

**Sample pitch offset (`2d5f832`).** The creator wrote only the sample rate at [54-57] and left [58-59] at zero, so every sample not stored at exactly 44100 Hz played back transposed. Implemented as `round(768 * log2(rate / 44100))`, which reproduces your captured values: 22050 → -768 (exactly one octave), 11025 → -1536, 44100 → 0, 48000 → +94.

Worth recording why this survived my testing: the detector reads the rate back from the informational field at [54-57], which was written correctly all along, so a detector → creator → detector round-trip reproduced every rate exactly and looked clean. Only hardware - or your capture - would have shown it.

**CD-ROM image directory entry (`c40b7ee`).** Same derivation and so the same bug: the bytes used in the last block came from `lastClusterBytes % 512` and was therefore 0 whenever the last cluster was an exact multiple of 512. It is now derived from the ceiling block count and is always 1..512. On a test image of seven files with sizes chosen around the boundary (512, 513, 1023, 1024, 100000, 524288, 524800), four of the seven entries were 0 before the change and none after, while the three that were already correct keep their exact values (1, 511, 160).

That does mean I have to correct what I wrote further up this thread: the images are byte-identical to the **pre-fix** builder. After `c40b7ee` they agree with yours again for every file whose last cluster is not an exact multiple of 512, and in that one case both now write the same corrected value.

**Docs (`e6b85b8`).** I mirrored the [58-59] field and the 1..512 rule into `documentation/design/E4B_FORMAT.md`, with the attribution for both.

Much appreciated - a port is only as good as the spec it comes from, and these two were not findable from this side.

douglas-carmichael:
Two more commits, from converting an E-mu Producer Series Volume 7 CD-ROM (2 discs, 45 banks, 2336 presets, 76057 zones) - the first real-world run against a full commercial library rather than single banks.

**`02ed659` - naming, and the placeholder preset**

Every bank of the commercial libraries ends with an unusable `Emu Systems 2000` preset whose only zone references a sample that is not in the bank. It was already dropped correctly, but it logged a `Sample with index 0 ... not found` error for each of its zones, once per bank. A zone with sample index 0 has no sample assigned (the indices are 1-based), so it is skipped silently now, and a genuinely dangling index is only reported for a preset which does contribute other zones - across both discs no surviving preset has one, so the log is clean.

The bigger issue was that the presets are named after their articulation (*Dark Tremolo*, *Long Release*, *Muted*, *Wide*) and never after the instrument, which only the bank name gives. 812 of the 2336 presets are not identifiable from their name alone, and the collisions are real: converting disc 1 produced `Dark Tremolo.xrni` and `Dark Tremolo (2).xrni` from two different banks. The bank name is prepended now unless the preset name already starts with it, comparing without spaces and punctuation so that `GreekBazoukiMute` and `Greek Bazouki XS` stay unchanged while `Dark Tremolo` becomes `Greek Bazouki - Dark Tremolo`. Each bank also gets a folder of its own when the folder structure is created.

Also in there: the name fields of the disk filesystem are padded with zero bytes rather than the spaces the bank chunks use, so bank names read out as `Greek Bazouki???` and would have carried those characters into the created names.

**`7ce000f` - the bank field, and the per-zone offsets**

Prepending is fine for a file name but wrong for a device with a small display, so it is a source option now (on by default) and the bank is additionally passed on as the description - which is where the formats with a bank field of their own already take it from, e.g. the Quantum/Iridium shows it next to the preset name.

Looking for anything else the format carries that was being dropped, a differential scan over all 76057 zones turned up three per-zone offset fields. mpc2emu writes zeros for those bytes and does not decode them, so this is new ground; the evidence is written up in `documentation/design/E4B_FORMAT.md`:

- **fine tune at `12:14`** - the high byte only ever holds `0x00` or `0xFF`, exactly matching the sign of the low byte, which identifies the pair as one signed big-endian value (range -59..+48). Measuring the recorded pitch of the samples against the root key of their zone shows the field to be a correction of that deviation: read as 1/64 semitones it drops the average pitch error of a bank from 12.3 to 7.7 cents, where reading it as cents only reaches 8.7 - and 1/64 semitone is the resolution the rest of the format uses for tuning anyway.
- **volume at `15`** - signed dB like the voice volume, small range (-17..+10), used by the mixed presets (*Percussion Section*) which balance their zones against each other.
- **panning at `16`** - full signed -64..+63 with the extremes dominating, and its users are exactly the stereo spreads (*Wide*, *Percussion Section*, *Congas/Bongos*) while the plain single instrument presets leave it at 0. This one is written as well now; panning has no voice level counterpart, so it was previously lost entirely.

The sample pitch offset is read too, as far as it detunes the sample beyond compensating its sample rate - which the sample data itself already does on this side. A deviation of more than a semitone means the bank does not maintain the field at all (banks written before the offset fix would look like that) and is ignored rather than transposing the sample.

A bank round-tripped through detector and creator keeps panning and volume byte-identical and is pitch-exact in the tuning; the tuning representation shifts because the sample pitch offset folds into the voice fine tune, the model having no sample level tuning of its own.

Feature matrix, changelog, README-FORMATS and the format spec are updated. Still no Emulator IV here, so the writing side remains hardware-untested.

douglas-carmichael:
One more commit, `c72bf38`, and a note on scope.

Converting a bank back to E4B showed that the bank prefix this PR adds to a preset name does not fit the format's own 16 character preset name field: `Greek Bazouki - Dark Tremolo` truncates to `Greek Bazouki -`, which collapsed the 20 distinct presets of a round-tripped bank into 6 names. Since the written file *is* the bank, its preset names do not need to repeat it - and could not, the bank alone fills the field. The bank is now dropped from the preset name when writing, which restores all 20, and the file name of the bank keeps the full name.

The same question arises for every destination whose name field is short, and for the Waldorf Quantum/Iridium, which has a Bank field of its own. That is a cross-format concern rather than an E-mu one, so it is **#228** and not part of this PR. The two are independent and can be merged in either order: this one carries its own small `createPresetName` in `Emulator4Creator`, #228 adds a shared `AbstractCreator.createNameWithoutBank`. If both land, the former can simply be folded into the latter.

lentferj:
I was able to hardware confirm the "per sample zone" parameters (they are not offsets, but absolute values, though).

https://github.com/lentferj/mpc2emu/commit/140edf06e4657d86e623a3ed321d81b5b1994ab8

PR #221, opened 2026-07-24
Kurzweil K2x00: convert the program DSP parameters (envelopes, filter, velocity window)

douglas-carmichael:
Converts the DSP parameters of the program layers. The byte layouts are based on the hardware reverse-engineering of the [mpc2emu project](https://github.com/mpc2emu), which decoded them by editing single parameters of the K2000 ROM default program on a K2000R, disk-saving each variant and diffing the files; the encodings below were confirmed against the device display and recordings. The details are documented in *documentation/design/KURZWEIL_FORMAT.md*.

**Reading** (per program layer, applied to its zones):

* The layer velocity window (two packed 0..7 dynamic marks, the high mark stored inverted), intersected with the 8 dynamic levels of the keymap.
* The amplitude envelope (the 7-stage ENV segment, times on the non-linear editor grid) - unless the layer plays the 'natural' envelope of its samples, in which case the category default still applies.
* The filter: 1/2/4-pole low-pass, 2/4-pole bandpass, high-pass and notch, with the cutoff (signed semi-tone byte, 16 Hz - 25088 Hz) and the resonance (0.5 dB steps, maximum 24 dB).
* The filter envelope (ENV2) with its modulation depth (piece-wise linear through the measured calibration points 40 = 1200 cents and 127 = 10800 cents).

**Writing:**

* The layer now carries the global amplitude envelope, filter and filter envelope of the multi-sample.

**Fixed:**

* The keymap object ID was written into both keymap slots of the calibration segment (as KurzFiler does). The device then claims two keymaps per layer, which overflows its limit at 4 or more layers and mutes the whole program (hardware-confirmed by the mpc2emu project; real hardware-saved programs keep the second slot at 0). Only the confirmed slot (bytes 11/12) is written now; reading accepts both.

**Not converted** (noted in the design document): the LFO segments and the vibrato routing, since the internal model has no LFO representation, and the pitch envelope (ENV3), since the control source code which routes it to the pitch is not confirmed.

The design document, README-FORMATS, the changelog and the feature matrix are updated.

**Verification (software only - NOT yet tested on real hardware):**

* SFZ → KRZ → SFZ round trip: filter type, cutoff, resonance and sustain return exactly; envelope times quantize to the device grid; a byte-level dump of the written program matches the reverse-engineered layout field by field.
* The written files load cleanly in KurzFiler (headless load harness).
* The KurzFiler test corpus converts without regressions (21 conversions, no errors, same as before).
* Factory banks (e.g. bassman.krz) read without false filters or envelopes; programs in 'natural' envelope mode still fall back to the category default.


douglas-carmichael:
Fixed the envelope issue reported with the FM bass bank. Two things were wrong for those programs:

**Unused envelope stages.** A stage whose time and level are both zero is unused on the device and keeps the level of the previous stage. The 'solid dx2' programs have `Att1 = (0s, 100%)`, an *unused* decay stage and `Rel1 = (10s, 0%)`: the envelope sustains at 100% and the 10s release just lets the naturally decaying FM sample ring out after note-off. The reader took the decay stage literally (decay to 0% in 0s), which produced the silent preset with only the release set. The reader now applies the unused-stage rule, and the writer avoids emitting an ambiguous all-zero decay stage for a genuinely percussive source.

**Velocity → cutoff.** The same programs route control source 100 (attack velocity) to the filter frequency with a closed 292 Hz base cutoff, so without that modulation the conversion was also very dull. The F1 modulation slot is now converted in both directions: source 121 = filter envelope, source 100 = cutoff velocity modulation (read into the cutoff velocity modulator of the model, e.g. `fil_veltrack` in SFZ).

With the fix the bank converts to SFZ as `ampeg_sustain=100  ampeg_release=10  cutoff=293.66  fil_veltrack=1110` (bass) / `5517` (8ve), and a round trip back to .krz reproduces the original F1 routing bytes (source 100, depth 37) and the 10s release. The KurzFiler test corpus sweep is unchanged (21 conversions, no errors) and the written files still load in KurzFiler. The design document is updated with both findings.

git-moss:
@douglas-carmichael Could you please send a PR for the last fixes?

PR #222, opened 2026-07-24
E-mu Emulator IV: read banks directly from EOS CD-ROM and hard disk images (stacked on #220)

douglas-carmichael:
**Stacked on #220** - only the last commit ("read banks directly from EOS CD-ROM and hard disk images") is new; the earlier commits are PR #220.

The EOS samplers read no standard filesystem from CD-ROM at all - their media use E-mu's proprietary disk filesystem (the same family the emu3fs Linux kernel module covers). Most Emulator IV content in circulation exists as images of that filesystem: dumps of the commercial E-mu CD-ROM libraries and images for SCSI emulators like the ZuluSCSI.

This adds reading of such images (*.iso*, *.img*, *.hda*):
- via the **E4B source format** (the file endings were added there), and
- via the **ISO/IMG source format** (a new `EMU3` entry in the format identifier, dispatched to the E4B detector - the same pattern as the Ensoniq and Roland image handling). The *.hda* ending was added to the ISO detector for hard disk images.

All filesystem geometry (block layout, cluster size) is taken from the superblock, so both the CD-ROM variant and the hard disk variant are read with the same code. Files which are not Emulator IV banks - e.g. EIII banks on discs of the older samplers, which use the same filesystem - are skipped. The filesystem layout is documented in `documentation/design/E4B_FORMAT.md` (reverse-engineered by the mpc2emu project from commercial CD-ROMs and EOS-formatted disks, cross-checked against emu3fs).

Validated against images generated by the mpc2emu reference builder (the three Ian Wilson banks as a CD image and as a 1 GB hard disk image): all 198 presets read through both source formats, sample PCM byte-identical to reading the plain bank files.


douglas-carmichael:
Folded into #220 - the same two commits are now part of the main Emulator IV PR, so everything can be reviewed in one place.

PR #234, opened 2026-07-27
List the three E-mu Emulators together in the format menus

douglas-carmichael:
The source and destination menus follow the order in which the formats are registered in `ConverterBackend`, and the three E-mu Emulators ended up in two places once all of them had been merged:

```
Expert Sleepers Disting EX
E-mu Emulator IV              <- with the Disting EX
Elektron Tonverk Multisample
Elektron Tonverk Preset
E-mu Emulator III             <- behind the Elektron entries
E-mu Emulator X
Ensoniq EPS/EPS16+/ASR-10
```

Each of the three was added by a separate PR against a `main` which did not have the other two, so none of them could place its entry next to the others. They are now registered as one block, in the order of their names, at the position the Emulator IV already had:

```
Expert Sleepers Disting EX
E-mu Emulator III
E-mu Emulator IV
E-mu Emulator X
Elektron Tonverk Multisample
Elektron Tonverk Preset
```

Registration order only - four moved lines in each of the two lists, no behaviour change.


git-moss:
Just did this as well :-)

PR #235, opened 2026-07-27
Repair the E-mu Emulator rows of the feature matrix

douglas-carmichael:
The three Emulator formats were merged into the sheet one after the other, and every one of those merges appended the new format's cells to the **end of the existing row** rather than adding a row pair of its own. The corruption is visible in the history:

| after | row 25 | cells |
|---|---|---|
| #220 Emulator IV | `E-mu Emulator IV` / Read | 70 |
| #229 Emulator X | `E-mu Emulator X` / Read | 135 |
| #230 Emulator III | `E-mu Emulator III / IIIX / ESI` / Read | 170 |

So on current `main` one row holds three formats side by side - Emulator III in columns 0-6, then Emulator X from column 9, then Emulator IV from column 14. Only the Emulator III has a name, and its own columns are overwritten from column 9 on; the Emulator IV and Emulator X have no rows of their own at all.

This restores them from the last revision in which each format still had its own pair, so the sheet now has six rows in the order of their names:

```
E-mu Emulator III / IIIX / ESI   Read / Write
E-mu Emulator IV                 Read / Write
E-mu Emulator X                  Read / Write
```

Each is the regular width again, the `TODO Check for support` markers are kept, and no other row of the sheet is touched.

Worth knowing for the next format merge: a text union is not a safe conflict resolution for this file. Two branches adding rows at the same anchor produce exactly this - the cells are concatenated into one row instead of the rows being stacked. Taking one side and re-adding the other format's rows afterwards avoids it.

### DIFF
--- a/documentation/SupportedFeaturesSampleFormats.fods
+++ b/documentation/SupportedFeaturesSampleFormats.fods
@@ -3348,588 +3348,437 @@
      <table:table-cell table:style-name="ce11" table:number-columns-repeated="2"/>
      <table:table-cell table:style-name="Default" table:number-columns-repeated="16315"/>
     </table:table-row>
-    <table:table-row table:style-name="ro10">
-     <table:table-cell table:style-name="ce2" office:value-type="string" calcext:value-type="string" table:number-columns-spanned="1" table:number-rows-spanned="2">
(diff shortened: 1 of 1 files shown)


git-moss:
Thanks!

Issue #23, opened 2023-09-24
Add support for Blackbox autosampler format


Zaat:
Feature request:

Be able to convert from (maybe to also) the Blackbox (1010music) autosampler which is a folder of WAV files. Although I think you could do it through the WAV file format conversion, it wasn’t straightforward for me. 

It’d be nice to have a template for it so you could simply select it for conversion. Despite it being fairly straightforward it’d be straightforward to implement and less work on the user part. 

The multi samples are stored in a folder per instrument and look like this:

![IMG_8241](https://github.com/git-moss/ConvertWithMoss/assets/16344035/23313ef3-9412-4846-be97-ba4559c6c2fa)

The first value is velocity and the second the MIDI number.

BTW, If there is a workaround for now does anyone have an idea how exactly with this MIDI note number? Thanks.

git-moss:
Don't think that this will happen (don't have a Blackbox) but will leave it open as a reminder.

benzkji:
@Zaat aren't there some .xml files lying around? They would contain the rest of the information, and would (hopefully) be quite easy to read. Will dig in if having some time...

git-moss:
Support was added in ConvertWithMoss 7.4.

PR #237, opened 2026-07-27
E-mu Emulator III: read the floppy disk sets of the EIIIX and ESI samplers

douglas-carmichael:
This makes the 1.44 MB floppy images of the EIIIX and ESI samplers readable, including banks which span several disks - the content of the recently shared TestFiles.zip (Emulator3X.zip and ESI-32.zip). Those images were already picked up by the EIII detector since #231 added the .img ending (they start with the same bank identifiers), but every one of them failed with thousands of 'Sample N ... is malformed' messages and 'contains no convertible presets'.

The reason is that the floppies are not bank files. They are a big-endian dump of the memory of the sampler, split across disks, with a 512 byte disk header on every disk (identifier, bank name, disk N of M):

* The address tables hold memory addresses instead of file offsets.
* The 92 byte sample headers are not stored in front of their sample data but in a separate table of 1000 slots, with positions relative to the channel data and their own option flag layout (channel presence, loop and loop-in-release sit in different bits than in a bank file; the fields of a channel which is not flagged hold stale values of deleted allocations).
* The sample data is stored per channel, addressed with bit 26 as a memory-bank flag which has to be masked off, and the 16 bit values are big-endian.
* The flag byte of a zone is stored with its bits mirrored. The 'Stereo Grand' library is the proof: all of its zones carry 0x80, which as a file flag would mute the right channel of a stereo piano.
* The EIIIX orders the regions record / sample data / header table, the ESI samplers record / header table / sample data.
* The loop end already points behind the last loop frame, where a bank file stores the frame before the last one.

The detector now recognizes a floppy disk (raw floppy size plus sane big-endian disk numbers, which a bank file cannot have), collects the other disks of the set from the same folder - by the bank name of their disk headers, since the file names of the collections in the wild number their disks in different ways - and converts the assembled payload into a little-endian bank which is then parsed by the existing code. Continuation disks are skipped with a log message, a missing disk cancels the set (same messages as the Roland S-7xx diskette handling). Floppy sets of the original Emulator III are not supported; no set to verify against has surfaced.

Along the way this fixes a preset-link misread which also affects bank files and CD-ROM images: the link at offset 49 is a single byte followed by a separate parameter at offset 50, not the 16 bit value emu3bm reads. A preset which sets both (e.g. 'Stick Combo III' of the 'Chapman Stick' bank) had an out-of-range looking link and lost the preset which is layered on top of it.

Verification:

* All 10 floppy sets of the test files and the Sweetwater collections read cleanly: 631 presets, 806 samples, no errors (previously 0 presets and 22817 malformed-sample messages). Zone resolution was cross-checked with an independent Python parser: 11634 of 11636 zone references resolve, the remaining 2 dangle in the source bank itself.
* Sample correctness: mono and stereo PCM byte-compared against the independent parser; stereo confirmed by the channel correlation of the samples (0.95 for the left/right pair of a stereo drum, -0.28 for a shifted control); loop seams measured over all looped samples (90.6% land under 5% of the peak with the chosen loop-end convention, the same quality as the bank-file reader after its loop fix).
* No regression on the EIIIX/ESI/Formula-4000 library CD-ROMs: the analysis output is identical to main except for five presets like 'Pad Link 1' and 'Link #023', which are link layers that the link fix now correctly folds into their parent presets instead of converting standalone.

The design document gained a 'Floppy disk sets' chapter with the full layout.


urbient:
you two did gods work here !

PR #238, opened 2026-07-27
Add a Batch toggle to convert one single file instead of a whole folder

douglas-carmichael:
A folder often contains many files of the source format of which only some should be converted, e.g. a couple of EXS programs or sampler disk images. Until now the whole folder had to be converted.

## User interface

Next to the source **Folder** button there is now a **Files** button, which opens a multi-selection file dialog filtered by the file endings of the selected source format.

The path field keeps showing the folder, the selection is displayed below it (*"Converting only 3 selected files of this folder"*) and can be discarded with **Clear**. Changing the source path - by typing, by picking from the history or by selecting a folder - discards the selection as well, so the displayed path never contradicts what is actually converted. Entering the path of a file instead of a folder converts that one file.

## Command line

One or more source files can be given instead of the source folder, which is most useful with a wildcard:

```
ConvertWithMoss -s exs24 -d 1010music MySampler/Piano*.exs Output
```

The last given path stays the destination folder, therefore the previous `SOURCE_FOLDER DESTINATION_FOLDER` form is unchanged.

## Semantics

A selection only replaces the file search, it is **not** a new grouping mode: each file is converted on its own, exactly like it would be as part of a full folder conversion, and the destination type still decides whether the results become single presets or one library. The Sample Files source keeps combining the files - its contract is that all sample files of a folder form one multi-sample - the selection just states which of them take part.

Sub-folders are not searched when files are given, since the selection already states exactly what to convert.

## Implementation

`IDetector.detect` gained a `List<File> sourceFiles` parameter; an empty list is the unchanged recursive folder search. The source folder stays a real folder in the detectors - the parent folder of the files - since it is the reference for all sub-path calculations, which is why nothing changes for the existing conversions. Detectors which do not process their source files individually can overwrite the new `detectFiles(List<File>)` hook, which only the Sample Files detector does.

## Testing

Tested with EXS24 sources against the Bitwig multisample destination, in the user interface and on the command line:

* A full folder conversion detects and writes the same files as before.
* A selection converts exactly the selected files, and the written `.multisample` files are **byte-identical** to the ones of the full folder conversion.
* A selected file which does not match the source format is skipped with a readable message naming the expected file endings.
* The Sample Files source combines a selection of AIFF files into one multi-sample, like it does for a folder.

---
**Changelog entry** (all entries of the open PR batch are collected in #283, so the PRs themselves do not touch the changelog and merge conflict-free):
```
* User Interface
  * New: Next to the *Select* button of the source there is now a *Batch* toggle. It is on by default and everything works as before: *Select* picks the source folder and all of its files of the source format are converted. Switching it off converts one single file instead - *Select* then opens a file dialog whose filter box lists every source format with its file endings, the current one preselected, and the source format is set from the picked file afterwards (a switched filter wins, otherwise the file ending decides; disk images go to the generic ISO/IMG format). Each mode keeps a path history of its own, which is exchanged along with the entered path when the toggle is used.
* Command Line Interface
  * New: One or more source files can be given instead of the source folder, which converts only these files (e.g. `ConvertWithMoss -s exs24 -d 1010music MySampler/Piano*.exs Output`). The last given path stays the destination folder.
```


git-moss:
@douglas-carmichael Hi Douglas, good idea. I think the user experience could be improved a bit:
- Keep the old layout. Add a toggle button after the Select-button labeled "Batch".
- Batch is enabled by default and everything acts like before
- If Batch is off, the Select button opens the file select dialog
- A separate history of single files should be kept as well
- If Batch is toggled, the history lists are exchanged and the last selection is set in the text field. If single files mode is selected, the source format is automatically selected based on the file ending (IMG, ISO, etc. endings select the Generic ISO/IMG detector)
- The file selection dialog should have all formats with their endings in the filter box, the current one is pre-selected
- If the format is switched in the file dialog it needs to be adjusted in the list after closing as well

douglas-carmichael:
Reworked to your design — the layout is back to what it was, with a **Batch** toggle next to *Select*.

* Batch is on by default and nothing changes: *Select* picks the source folder and all of its files of the source format are converted.
* Switched off, *Select* opens a file dialog for one single file. Its filter box lists **every** source format with its file endings and preselects the current one.
* After closing, the source format follows what was picked: a filter which was switched in the dialog wins, otherwise the file ending decides. The disk-image endings (ISO, IMG, HDA) go to the generic **ISO/IMG** format, since the E-mu formats claim the same endings and the generic one hands the image on to whichever of them can read it. An ending no format claims leaves the list alone. Verified: `.iso`/`.img`/`.hda` → ISO/IMG file, `.exs` → Logic EXS24, `.e4b` → E-mu Emulator IV, `.dspreset`/`.dslibrary` → DecentSampler, `.sf2` → SoundFont 2, `.nki` → NI Kontakt.
* Each mode keeps a path history of its own. Toggling exchanges the two and puts the last path of the mode being entered into the text field; both are stored in the configuration.

The multi-file selection is gone with it, and so are its status line, its *Clear* button and the two messages about a selection whose files disappeared — a single file needs none of that, and entering a file path by hand was and stays the same thing as picking one. That also removes the extra row from the layout, which is what you asked for.

One note on the stack: #241 builds its *Contents...* selection on the multi-file selection this replaces, so I will rework it onto the Batch design before it is worth another look.

PR #241, opened 2026-07-28
Select individual presets of a bank or a disk image to convert (stacked on #238)

douglas-carmichael:
**Stacked on #238** - its commit is the first one here, so please look at the second commit only. Rebasing once #238 is decided is no problem.

One source file can contain far more presets than the ones which should be converted. A CD-ROM image of an E-mu sampler holds thousands of them in dozens of banks, and the same is true for the disk images of the Akai, Ensoniq and Roland samplers, for SoundFont 2 banks and for the libraries of several plug-ins. Until now such a source could only be converted as a whole.

For several of these formats ConvertWithMoss is the only practical way to get at the content on a current machine at all - the samplers are decades old and the material only becomes playable by converting it. Having to convert a whole 700 MB CD-ROM image to get at three presets is a real obstacle to that.

## What it does

A source format where one file holds several presets has a **Contents...** button. It reads the source without writing anything and shows everything it found as a tree of the file and of the containers inside of it:

```
EmuCD1.iso01.iso
  Greek Bazouki
    Greek Bazouki XS      (88 zones, C-2-C6)
    Mute/Pluck/Trem      (133 zones, C-2-C6, Pluck)
  African Melodic
    Bamboo Layers 3       (24 zones, C1-C5, Ethnic)
```

Each entry shows its number of zones, its key range and its category, which is what one needs to explore a bank which is not known yet. Ticking a bank ticks all of its presets. Only the ticked presets are converted; how many of them are selected is shown below the source path together with a *Clear* button, and changing the source path, the source format or the destination type discards the selection since those change what is detected.

## How it is implemented

The selection is applied in `ConverterBackend`, where every detector delivers its sources, so it works for **all formats without any format specific code**. Verified with E-mu CD-ROM images and, as a completely unrelated second format, with Logic EXS24.

A source is identified by **its index inside of its file**. That is deterministic and therefore addresses the same source in the following conversion run. Names cannot be used, since a bank may well contain several presets with the same name. Counting the index per file - instead of over the whole run - additionally allows to skip the files which contain no selected preset without shifting the indices of the others, so those files are not even opened.

## Cost of the contents run

The run currently parses the samples although only the names and the zone information are shown, which is the one thing I am not happy with:

| | Contents | Convert 2 presets |
| --- | --- | --- |
| Both E-mu images (1.1 GB, 2339 presets) | 36 s | 36 s |
| ... after skipping files without a selection | 36 s | 18 s |
| One image selected with *Files* (#238) first | **9.9 s** | **10 s** |

Selecting the files first is therefore the practical answer today. A real fix is possible: the preset names come from `decodeName (data, presetOffset)` and the zones from the preset structure, both of which are read from the bank headers, so a detection which stops before the samples are decoded would make this nearly instant. That needs a way to tell a detector that only the information about the sources is wanted, which I would rather do as a follow-up once the approach here is settled.

## Questions

1. `ConverterBackend.MULTI_PRESET_FORMATS` lists the formats which get the button - 25 of the 50 source formats. Each entry was checked against what `readPresetFile` of that detector can actually return, so e.g. Akai MPC modern, Omnisphere, Roland MV-8000 and ZEN-Core are out because they return a single preset per file, while 1010music blackbox/bento are in because they return one multi-sample per pad. I kept the classification in one place to have it reviewable at a glance, but it could just as well be a method of the detectors, like `supportsPerformances ()`. Which do you prefer?
2. Should the tree start collapsed? Two CD-ROM images give 2339 entries.

Not done yet: auditioning a preset before converting it, which would be the natural next step for exploring an unknown bank but is a scope question of its own.

---
**Changelog entry** (all entries of the open PR batch are collected in #283, so the PRs themselves do not touch the changelog and merge conflict-free):
```
* User Interface
  * New: A source format where one file contains several presets - a bank, a disk image or a library - has a *Contents...* button, which shows all presets found in the source as a tree of their file and their containers. Each entry shows its number of zones, its key range and its category, so that an unknown bank can be explored, and only the ticked presets are converted. Switching *Batch* off first narrows it down to one single file, which keeps it fast on a folder full of banks.
```


douglas-carmichael:
Reworked onto the *Batch* design of #238.

The selection line below the source path now only ever reports presets — "Converting only 5 of 229 found presets." — since the multi-file selection it used to also describe is gone. `IDS_MAIN_SELECTION`, `..._FILE` and `..._FILES` collapsed into a single `IDS_MAIN_SELECTION_PRESETS`, and `clearSelection` only has the preset selection left to clear.

The contents run reads the source exactly the way the conversion does, which is now shared: `applySourcePath` turns the entered path into a folder or into one single file, and both `openContents` and `verifyFolders` call it. So switching *Batch* off narrows the contents run to that one file, which is what makes it usable on a folder full of disk images — measured on two E-mu CD-ROMs:

| | presets found | time |
|---|---|---|
| Batch on, whole folder | 2885 | 14.4 s |
| Batch off, one image | 229 | 0.23 s |

The selection is discarded whenever the source changes — editing the path, picking a folder or file, **toggling Batch**, or switching the source format or destination type — because it addresses presets by their position in the detection run of exactly that source.

Verified end to end: with two disc images in a folder, ticking one preset in each converts exactly those two; a DecentSampler `.dslibrary` of three presets lists all three. Also restored the CRLF line endings on `IDetector.java`, which had been converted to LF on this branch and showed up as a whole-file diff.

PR #245, opened 2026-07-28
E-mu Emulator III/IV: do not convert a parked filter cutoff into silence

douglas-carmichael:
The presets of the EOS libraries commonly park the filter cutoff at an end of its range and open it up again with a modulation cord. Only the filter envelope and the key tracking of such a cord have a counterpart in the model — velocity, the assignable MIDI controllers and the LFOs are lost — so what was left was a filter which removes the whole signal, and the preset converted to silence.

This is not a rare corner: measured over all 13 CD-ROMs of the E4 factory library (42935 voices, 38783 of them with a filter type the model can hold),

* **12306 voices sit parked at an end of the range**, and
* **every single one of the 11065 parked low-pass voices has a cord into the cutoff — no exceptions.**
* On *Producer Series Vol. 7 – Old World Instruments* **every** low-pass is parked: 33880 zones came out as a 4-pole low-pass at 57 Hz and not one at an audible frequency, so the whole CD-ROM converted to silence in every destination which honours filters.

**Emulator IV.** A voice whose static cutoff alone suppresses everything, and whose cutoff is modulated by a cord which cannot be converted, now gets no filter at all. That is much closer to the original than the silence it produced before. A voice whose filter envelope pulls the cutoff *away* from that end keeps its filter, since the model opens it up as well — 1866 voices take that path, 10390 lose their filter and 50 are parked with no modulation at all and stay as they are. This is the same reasoning `EmulatorXDetector.createFilter` already applies to the EXB banks, generalised to the filter type and to the direction the envelope moves.

The reported case is `3DPads - 3DreamSwells` on *Vol. 5 – 3-D Audio Collection*: a 2-pole high-pass parked at 14.5 kHz with four cords into the cutoff, of which only the filter envelope is convertible.

**Emulator III.** The cutoff parameter reaches up to 74 kHz and the banks switch the filter off by parking it anywhere above the audible range, but only the single value `0xFF` was taken as that bypass state. `0xEF` alone — the value `DEFAULT_CUTOFF` already documents as the fully open filter — accounts for 30411 of the 144933 zones of the EIIIX library. The frequency now decides instead of one value, which drops about 3200 inaudible low-pass filters across the library CD-ROMs and leaves every audible one in place.

**Verification.** Both readers were run over the full corpora (13 E4 CD-ROMs, 8 EIIIX/ESI CD-ROMs) with no errors; the byte-level statistics above come from those runs. Before/after counts of the emitted SFZ filter opcodes confirm the change only removes filters which cannot be heard — *3-D Audio Collection* still writes 547 of them, and the four EIIIX discs which are affected go 19261→17698, 22769→21419, 11662→11634 and 28617→28346.

Two things I checked and left alone, in case they come up:

* The voice offsets were never the problem. A histogram of `vpar[58]` over all 42935 voices lands exactly on the documented `group|variant` scheme (`0x00`–`0x02`, `0x08`–`0x09`, `0x10`–`0x12`, `0x20`–`0x22`, `0x40`–`0x42`, `0x48`, `0x50`–`0x51`, `0x60`–`0x62`, `0x68`, plus `0x7F` = bypass) and never on a plain index, and `vpar[60]` spikes at 255, which confirms it is the cutoff.
* A band-rejection at an end of its range passes everything rather than nothing, so it is excluded from the test.

No conflict with #242, which touches the envelope and loop reading of the same detector; the two can merge in either order.


douglas-carmichael:
Added a second commit after finding the EIIIX operations manual, which settles a question the first commit had to work around.

The manual documents the filter Tracking parameter as running from -2.00 to +2.00 and says of 0.00 that "the filter cutoff will not be affected by the keyboard pitch"; its own walkthrough of the VCF page shows `Tracking: +1.00`, which is the byte 0x40. So 0x40 is full tracking, and `NO_VCF_TRACKING = 0x40` was not the neutral value it claimed to be — the creator writes it for a zone which has no filter.

The consequence was a round trip which did not hold: a source without a filter, written to an E3X bank and read back, returned with a 45 kHz low-pass tracking the keyboard at 1200 cents per octave (the wide-open cutoff alone would have been recognised as the bypass state by the first commit, but the tracking kept it alive). A zone without a filter now writes no tracking, and the same two-zone source returns with both zones and no filter.

Only the creator uses that constant, so reading is untouched — *Elements of Sound 1MB* still yields exactly the same 17698 filters before and after.

The same manual confirms two things the first commit relies on: the cutoff display reaches 74040 Hz, matching the last entry of the conversion table, and 0.00 tracking is described as what "generally gives the most realistic synthesizer and instrument sounds", so writing it is not an odd choice for a filterless zone.

PR #288, opened 2026-07-30
Center presets rooted far off the keyboard middle and audition at a sample root

douglas-carmichael:
You are the engineer in a session at Hansa in Berlin - or in New York, Los Angeles, Nashville, London, Paris. The composer hands you their vintage E-mu library - twenty years of their sound - and asks for the vocals on the keyboard. You convert the bank, load it, play middle C: a slow, groaning moo comes out. The room goes quiet, and the composer is watching. "The source maps it that way" is even true - vintage libraries root vocal phrases, talk box, sound effects and single-sample basses an octave or more above the middle of the keyboard, and a faithful conversion keeps exactly that - but true does not help you right now. You are the person who could not deliver, and that costs the next call, not just this moment. Nothing was broken. Nobody in the room can tell. Measured on E-mu Producer Series Vol. 8 (CD 2): 949 of its 1046 presets are mapped exactly like this, rooted at MIDI 72.

After the discussion below this PR changes no conversion output at all. It closes the information gap instead, fully automatically:

* The *Play* button of the *Contents...* dialog renders its note at a sample root close to the covered middle instead of the plain middle of the covered range, so a phrase preset auditions as it was recorded (before, the preset which started all this previewed 8 semitones slow).
* When a conversion leaves a preset whose samples are rooted an octave or more from the keyboard middle, one log line names that root and points at the existing Transpose processing option: `Note: 'Female Vocals - Oh' plays its samples as recorded at C4 (MIDI 72) - around the middle of the keyboard it sounds strongly transposed, exactly as the source maps it. The Transpose processing option can move it.` The line appears in the Analyse run as well. A preset with more than two distinct roots is a crafted multi-sample whose placement is a design decision and gets no note.

The line earns its place: while building this PR I lost half a day hunting a pitch bug in a destination synth that does not exist - the converted preset was faithful all along, I was playing it two octaves below its root. That happened to the person who wrote the E4B support, with the source code open. One log line ends that hunt before it starts.

(The octave-centering and its setting from the earlier revisions of this PR are withdrawn per the discussion; my fork keeps them for studio use.)

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -10,9 +10,11 @@
 * New: Added support for the E-mu Emulator X format (EXB banks with their EBL sample pool).
 * New: Added support for the Roland S-550 CD-ROM format.
 * New: Added support for the Roland SP-404MK2 format (reading and writing of projects; each bank of pads becomes a multi-sample).
+* New: When a converted preset plays its samples an octave or more from the middle of the keyboard (common for vintage phrase and vocal presets, which were triggered from drum machines rather than played on keys), a log line names the root it plays at, so the faithful mapping is not mistaken for a conversion error. The existing Transpose processing option can move it.
 * User Interface
   * New: There is now a toggle switch before the source field to switch between batch conversion (as it worked before) or only picking a single file to convert. Each mode keeps a history of its own, which is exchanged along with the entered path when the toggle is used.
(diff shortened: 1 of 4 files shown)


git-moss:
@douglas-carmichael Nice introduction story ;-) Applying it to the playback is definitively helpful. But I am a bit unsure about the option. I guess the reason why it is low, is that back then they sounds were triggered by drum machines in the drum note range. Therefore, the option should be off by default. Since I want to keep CWM as automatic as possible, I am not a fan of too many options (like in AWave or Translator) and would prefer not to have it at all.

douglas-carmichael:
You are right about the drum machine history - it matches the measurements, and it is also why the centering already skipped every preset with more than two distinct roots. And I follow you on the options philosophy, so I have cut the PR down to what is fully automatic: the *Play* preview at a sample root stays, and instead of the centering there is now a single log line when a conversion leaves a preset rooted an octave or more from the keyboard middle. It names the root and points at the existing Transpose processing option; crafted multi-samples (more than two distinct roots) stay silent. No option, no changed output.

The reason I kept that line: the issue is real even for experienced users. While building this PR I spent half a day hunting a pitch bug in a destination synth that does not exist - the converted preset was faithful, I was simply playing it two octaves below its root. On Producer Series Vol. 8, 949 of its 1046 presets are mapped like this. One line in the log ends that confusion before it starts. The centering itself lives on in my fork only.

PR #289, opened 2026-07-30
E-mu EIII: stereo samples stored right-channel-first, the single-byte zone sample index of the Emulator III, and deleted-slot references

douglas-carmichael:
This resolves the sample-lookup errors on the first E-mu library CD (`Vol. 1 - Emulator Standards`, the archive.org image, MD5 `64dd1176a6ffd142bf8a9e5911f079b4`). The CD exposed three separate defects; all its 44 banks now read without a single error and the log matches what the sampler itself would find.

**1. Stereo samples are stored in either channel order.** The reader assumed the left channel always starts at byte 92 and computed the frame count from the end of the header. The CD stores 16 stereo samples right-channel-first (`Stereo Strings` slots 1-8, `4 Piece Horns 8M` slots 8-11/14-16, `Pop Ahs` slot 1): their header reads `startRight=92` and `startLeft` behind the right channel. Two of them happened to overrun the bank and were rejected as malformed - these are the `Stereo Strings' ... malformed: 1, 6` errors and the fourteen follow-up `not in the bank` errors of the string presets. The other fourteen were worse: they read "successfully" at **double length with mixed-up channels** and converted as garbage without any message. Each channel is now read from its own start/end fields. Verified by converting the CD and byte-comparing the WAV PCM of previously-rejected, previously-misread and untouched control samples against an independent extraction of the bank data - all identical, and the zone-level channel-mute path (`Pop Ahs - Chorus`) now yields exactly the right channel.

**2. The original Emulator III reads only the low byte of the zone sample number.** An `EMULATOR THREE` bank holds at most 99 samples, and the byte behind the number carries unrelated values on this CD (0x01..0x0C). `Groupo Sinfonia`'s `String Orchestra` is the cleanest proof: its zones store 0x0A02..0x0A09 with original keys E2, A#2, E3, C4, G4, E5, A#5, E6 - and samples 2..9 of the bank are named `Strings E2` ... `Strings E6`, an 8-for-8 match at the low byte. The heuristic high-byte repair already recovered most of these (`Full Arco String`, `Dance Club`, `Groupo Sinfonia` - the three "repaired" log lines), but it left `Slow Strings` reporting the impossible number 3082 because that reference's low-byte slot is deleted. Emulator III banks now mask the index to one byte, which resolves all of them deterministically; the repair heuristics remain for the u16 formats (EIIIX/ESI), which are untouched.

**3. References to deleted sample slots are data, not read errors.** The remaining reports (`Trombone 2` slot 55, `Groupo Sinfonia` slot 10, `Dance Club` slot 28, ...) point at empty sample-table slots - what deleting a sample leaves behind; the CD ships 23 such presets. The device finds nothing at such a slot either and keeps those keys silent (mostly a single top key, e.g. D5 of the trombone presets), which is why Awave shows no errors there: it drops them quietly. They are now reported once per preset as a plain note - "... reference deleted sample slot(s) and were skipped - these keys are silent on the device as well" - and only a reference that cannot be explained by the stored data remains a red error.

Before/after on the CD: 37 red error lines -> 0, with 23 neutral notes; `Stereo Strings` reads 14 of 14 samples instead of 12; every other bank's preset and sample counts are unchanged. The EIIIX/ESI floppy sets of the test files still read cleanly (their one stale reference, `Krz.Touch Orch.` slot 231, correctly reclassifies from error to note). `EIII_FORMAT.md` documents the channel-order and single-byte findings.

One thing I noticed while adding the changelog entries: in the current `CHANGELOG.md` the `* E-mu Emulator III` heading is glued to the end of the DLS line above it (missing newline), and the E-mu Emulator III/IV entries that were merged earlier (the malformed-report batching, the filter bypass counts, the truncated-sample-number repair, vibrato/chorus, both Emulator IV fixes, and the Tonverk/Ensoniq/Mirage/ISO/Korg sections) are no longer in the file - they appear to have been lost in a recent docs edit. This PR restores the newline so the section renders again, but only adds its own three entries; the lost lines are probably worth restoring from git history separately.


git-moss:
@douglas-carmichael Thanks for the quick fix! Works great now! One last thing: it still shows the error:
"The bank 'E-mu Banks 1-44' contains no convertible presets. Skipped."
Looking at the Awave output it seems to me that "E-mu Banks 1-44" is simply the name of the root? So, I suggest to either log it as the CD-name or simply ignore it.

Regarding the changelog, I removed the EIII fixes intentionally, since so far the EIII support is not released and fixes during the implementation/test phase are not helpful to the end users.

douglas-carmichael:
Right - 'E-mu Banks 1-44' is the name of the volume. The CD stores it as an additional *empty* bank (1867 bytes: the header, the two empty address tables and the filler byte, nothing else) and the sampler and Awave show that bank's name as the name of the disk. The detector now recognizes an empty bank inside a disk image as this label and logs "The disk is named 'E-mu Banks 1-44'." instead of the error; the bank count of the image line drops from 45 to 44 accordingly. The convention is documented in EIII_FORMAT.md.

The three CHANGELOG entries are removed again - understood about the unreleased format.

git-moss:
Perfect, thanks!

douglas-carmichael:
Ran the new code over the Classic Series ISOs from archive.org (the set matches your file names, so no CDs needed) and found three more root causes behind the log. All are fixed and verified locally on Vols. 1, 2, 9, 12 (CD 1), 13 and 16 - those six now read with zero errors.

**1. The remaining 'contains no convertible presets' warnings are more label banks.** The CDs store not only the disk name but the section titles of the bank list as empty banks - 'INSTRUMENTS', 'EFFECTS', 'GROOVE LOOPS', 'Credits', a 'Rev A' revision marker (Vol. 12 CD 1 alone carries six of them). Many of these stubs also hold one to three presets which map no key at all, which is why the first version of the emptiness check missed them. Emptiness is now decided by playable presets and samples, and since a single 'disk name' would be wrong for a disk with six labels, the message is now: *The empty bank 'X' is a label which names the disk or a section of the bank list. Skipped.*

**2. The '92/24 malformed samples' and the Trombone/'DAVE W KIT1' errors came from the sampler's operating system.** 'E3 Main Code' and 'E3X Main Code' are stored with the directory entry type 0x80 (system file; banks are 0x81 - emu3fs calls it EMU3_FTYPE_SYS), and on Vols. 13/16 their content is a memory dump which itself begins with a valid 'EMULATOR 3X' identifier and stale preset structures, so they parsed as banks full of garbage. The filesystem reader now skips system files.

**3. The remaining missing-sample errors are dangling preset links.** Banks like 'CONCRT PERC2   X' build their maps as a link chain of single-instrument presets, and the last link of a chain points at a slot which holds no preset. An empty slot stores the same address as its successor, so following the link read note zones and zones from behind the last preset. Simulating that read reproduces the reported numbers byte-for-byte: 'CONCRT PERC2 lnk' and 'CROTLE kyb' both link to the empty slots 126/127, which share one address - hence both reported the identical list 1490, 3511, 4105, 10730, 15604, 15614, 16329, 16356 ('Power Combo LW X' of Vol. 9 is the same pattern). A chain now ends at the first empty slot, which is what the device does; the preset counts of the affected banks are unchanged (4 / 2 / 9).

Across the six volumes here the log now shows only the label notes and the neutral deleted-slot notes: Vol. 12 CD 1 = 39 banks + 6 labels, Vol. 13 = 52 banks (was 54 incl. the OS file and the stub), Vol. 16 = 45 (was 48). The Formula 4000 'Credits'/'Credits & Info' stubs and the other volumes in your log match these same three patterns, so a re-run over the full set should come back clean.

PR #296, opened 2026-07-31
Fail with an error message when the destination cannot write the requested library or performance

douglas-carmichael:
Creating a library or a performance is an empty default operation for a destination which does not support it, so everything that was detected got collected and then silently dropped: the run reported "Finished." and left an empty output folder. Converting 175 Elektron Tonverk presets to Waldorf QPAT with `-l MyLib` wrote nothing, as did Kontakt multis to QPAT with `-t performance`. The command line accepts `-l` and `-t performance` for every destination, so this is easy to hit.

The command line now fails with an error message before the detection starts (updated after review - the first version fell back to writing single presets):

* `SFZ cannot write a library which contains several presets. Remove the option '-l' or choose a destination format which supports libraries.`
* `SFZ cannot write performances. Remove the option '-t performance' or choose a destination format which supports performances.`

The user interface is not affected, since it only offers the output types which the destination supports. Destinations which do support libraries or performances are untouched: the same source to a Synclavier V library still writes one synx file, and a plain preset conversion is unchanged.


git-moss:
Shouldn't that simply fail with an error message?

douglas-carmichael:
Agreed - `-l` is an explicit instruction, so reinterpreting it was the wrong call. It now fails with an error message before the detection starts (the capabilities are known upfront, so no work is wasted). The fallback is gone and `ConverterBackend` is back to stock; the check lives next to the other command-line validations.

PR #298, opened 2026-07-31
Arturia Synclavier V: option to write the samples into a sample pool folder

douglas-carmichael:
Adds a creator option (off by default) which writes the samples of each preset additionally as plain WAV files under `User/<preset>/` next to the written synx files, byte-identical to the embedded copies. Merging that `User` folder into the Arturia sample pool makes imported presets permanent, including for the standalone application. CLI parameter: `SynclavierVWriteSamplePool=1`.

Background - the reported behavior with converted synx files on Windows 11:

* The import only works when the plugin runs inside a DAW; as a standalone and via Analog Lab V it fails.
* Loading the imported preset in the standalone reports missing samples such as `C:/ProgramData/Arturia/Samples/Synclavier V/User/01W a realD50/01W a realD50_ms0_036_c1.wav`, and no `User` folder exists at that path.
* The plugin instance recalls the patch with its samples, but a search of the whole disk finds the sample files nowhere; only the preset file exists, at `C:\ProgramData\Arturia\Presets\Synclavier V\User\01W a realD50`.

All of this has one cause, reproduced and measured on macOS 2.13.4: the application's import loads the embedded samples but never copies them into its sample pool. The browser database receives the preset stripped of the appended sample archive (measured: exported file 168,882 bytes = database copy 55,495 + embedded WAV 113,392 - which is why only the preset file is found on the disk), and the embedded files are extracted to their `User/<preset>/...` reference path resolved against the **working directory of the host process** instead of the pool root. Inside a DAW that directory is writable, so the import succeeds and the WAVs land wherever the DAW's working directory points - Bitwig on macOS drops them under `~/Library/Logs/Bitwig/User/...`; on Windows they should be in a `User` folder near the DAW executable or its startup folder. A standalone launched normally has an unwritable working directory (System32 or Program Files), which matches the import failing exactly in the standalone contexts. Since the pool is never populated, a database preset opened after the importing session reports its samples as missing at the pool path. A genuine Synclavier V export imported on another machine runs into the same flaw, so this is an Arturia importer bug; the option makes converted files usable despite it.

The import behavior is documented as §4.1 of the format specification and in README-FORMATS.

Verified:
* SFZ → synx with the option writes `<out>/User/<preset>/<sample>.wav` byte-identical to the sample archive embedded in the preset; without the option the output is unchanged (exactly one synx file).
* macOS 2.13.4: with the extracted WAV placed at `Samples/Synclavier V/User/<preset>/`, a preset imported in an earlier session loads with its samples in the standalone.


douglas-carmichael:
Confirmed on macOS (2.13.4): with the WAV extracted from the synx placed at `Samples/Synclavier V/User/Female Vocals - Oh/Oh.wav`, a preset imported in an earlier session now loads with its samples in the standalone. Merging the written `User` folder into the pool is exactly that operation.

Issue #70, opened 2025-01-24
Wondering if support for .instr, .dwp, .dwb could be added


rickdtc:
Would be helpful a lot for FL Mobile users since desktop FL users already have a sf2 player.

It would be great if sf2 can be converted into multiple monolithic dwp and instr files!

You may be wondering, yes I would need a desktop to use this conversion tool, I have a 4gb i3 laptop which can not run any daw good, I have tried using FL Mobile alternatives with sf2 support, but if I could just use them directly in FLM then it would be better, will be way more than grateful if this was added! If possible ofc

git-moss:
I have currently no plans for this. Is there a documentation about these formats available?

rickdtc:
I tried to find some but I couldn't, I think these work simmilar to sf2 but I may be wrong

rickdtc:
Ok I found out that apparently [you can extract
.instr files with 7zip](https://www.reddit.com/r/FL_Studio/comments/secoxn/comment/huilrjh/?utm_source=share&utm_medium=mweb3x&utm_name=mweb3xcss&utm_term=1&utm_content=share_button)

I have not tested since I am currently away from my laptop, but here is a [sample .instr file](https://1drv.ms/u/c/235812e791d0cda0/EZRm7PaEYn5Aq_PFUtgDUsEBf7U9fX65IvgRqYjisRQhCw)

rickdtc:
https://github.com/ChaseDurand/DirectWave-to-Sampler

Found a project that might help for dwp

douglas-carmichael:
Status update: DirectWave support is implemented and opened as #301.

**Works now** - reading DWP programs (key/velocity ranges, root key, gain, panning, all five loop modes, amplitude envelope, filter), including the embedded samples of monolithic programs, and writing the folder layout which DirectWave loads directly and FL Studio Mobile imports (the DWP plus its samples as WAV files). A DWB bank or a plug-in sampling folder has its mapping reconstructed from the sample file names. So converting e.g. an SF2 into DirectWave programs works.

Since the format is undocumented it was reverse engineered; the specification is included in the PR as `documentation/design/DIRECTWAVE_DWP_FORMAT.md`. It was verified against the six DirectWave programs of the FL Studio Mobile factory data, and 509 programs converted from an E-mu bank were audited field by field. Written programs load in DirectWave (FL Studio 2026 on macOS).

[...]

PR #301, opened 2026-08-01
Add support for the FL Studio DirectWave format (#70)

douglas-carmichael:
DirectWave is the sampler of FL Studio; FL Studio Mobile plays its programs as well, which is what #70 asks for. The DWP format is not documented, so it was reverse engineered for this; the full specification is in `documentation/design/DIRECTWAVE_DWP_FORMAT.md`.

**Reading** a DWP program gives name, key and velocity ranges, root key, gain, panning, loop, amplitude envelope and filter per zone. All five DirectWave loop modes are converted, including *One-Shot*, *Sustained* (the model's loop-until-release) and *Bounce* (alternating). Monolithic programs are read as well - their samples sit in one block per zone as a FLAC stream. A DWB bank, or a folder written by the plug-in sampling feature, has its mapping reconstructed from the sample file names, which supports both the names DirectWave writes itself (`Piano 4xCycles_C5_100_2.wav`) and the Automap tokens of the manual (`sample_C4+KA3-F#4+V64-95.wav`).

**Writing always produces a monolithic program**: one self-contained `.dwp` per instrument which carries all of its samples as FLAC compressed audio, up to 24 bit. That is the form #70 asks for and it is what one converts for: a single file loads in FL Studio Desktop and in FL Studio Mobile, where a program plus a folder of loose WAV files would have to travel as two parts. The creator has no options.

How the undocumented parts were pinned down:

* The block structure, the zone mapping and every parameter block the converter does not fill come from a real monolithic FL Studio Desktop export, so a written program equals a real one byte for byte except for the six blocks which carry the converted content. Two fields are load-critical: the sample count in the header must match the number of sample blocks, and a sample must fill DirectWave's 512 frame audio blocks - the plug-in crashes with an access violation otherwise, which is why written audio is padded with silence to a full block.
* The filter type, LFO waveform and loop mode enumerations, and the modulation source and target names, are taken from the tables inside the plug-in binary, cross-checked against the menu definitions in its editor library (whose loop values match what the factory files contain).
* The position of the filter type field, the cutoff frequency law and the envelope time law were measured by playing purpose-built programs whose zones differ in exactly one knob, against reference zones whose behavior is baked into their audio. Cutoff is exponential from 45 Hz to 20 kHz, envelope times are `9.07 * position^3.96` seconds; both reproduce the measurements within a few percent. FL Studio ships its own value for the second one in `Data/System/FL Mobile/DirectWave.info`, `sqrt(sqrt(value/10))` = `10 * position^4`, which agrees with the measurement.

Reading was verified against 117 programs covering all three format versions: the six FL Studio Mobile factory programs, the 26 programs of the FL Studio legacy pack (their samples are Ogg Vorbis in a WAV container, which #303 adds) and 85 monolithic commercial programs. As a scale test an E-mu Emulator X bank of 509 presets was converted and all 509 written programs audited: no structural defects, 3598 loops and 1401 filters carried over.

Writing is verified on hardware. A 63 zone instrument with three velocity layers, converted from a Logic EXS24 auto-sampled instrument, loads and plays correctly in DirectWave (FL Studio 2026, macOS) and the same file loads and plays in FL Studio Mobile on iOS.

Not converted: trigger groups (only the first round-robin cycle is kept), the LFOs and the modulation matrix.


douglas-carmichael:
Update: monolithic programs are now read **and** written, so the whole of #70 is covered.

A monolithic program keeps the same structure and adds one block per sample container - tag `0x0206`, the last block before the container terminator - whose payload is the length of the audio, four unused bytes and then the audio as a **FLAC** stream. That is why the samples of a monolithic program are not found as separate files: they are compressed inside the program.

The tag was located in the block writer of the plug-in binary, where the payload length is a constant 8 bytes before the tag for every fixed-size block - so the blocks *without* such a constant are the variable-length ones, i.e. the sample name, the sample path and the otherwise unused 0x0205/0x0206. It was then confirmed on 85 monolithic programs of a commercial library: in all 288 sample containers the length field equals the payload size minus 8, the field at offset 4 is 0, and the block sits between the last 0x0204 block and the terminator. All 85 programs convert, 2040 samples extracted, none empty or unreadable.

Writing is a new *Monolithic file* option (off by default). A written program is structurally identical to one saved by DirectWave itself - same block order, same header fields, same audio block layout - and reading it back reproduces all zones.

The PR also adds the format version 0x24 of the legacy patch pack which ships with FL Studio (`Data/Patches/Packs/Legacy`), whose preamble is 4 bytes longer; the start of the block stream is now located rather than assumed. Those 26 programs are also the only files anywhere with an engaged filter - 111 low-pass zones - which independently confirm the filter type encoding. Their samples are WAV files with the Ogg Vorbis codec (format tag 0x674F) which the sample loader cannot decode, so their audio is not converted.

douglas-carmichael:
Verified in DirectWave: a monolithic program written by ConvertWithMoss loads and all of its zones play correctly (FL Studio 2026, macOS). Monolithic is therefore the default for the destination now - it is what #70 asks for, it is one file per instrument for the FL Studio Mobile use case, and it is smaller than the same instrument as a folder of WAV files because the embedded audio is FLAC compressed. The folder layout stays available as an option for the case that the samples should be editable.

Summary of what is verified on real files at this point:

* Reading: the six FL Studio Mobile factory programs, the 26 legacy programs which ship with FL Studio (format version 0x24), and 85 monolithic programs of a commercial library (2040 samples extracted, none empty or unreadable).
* Writing: the factory programs re-written and compared field by field, 509 programs converted from an E-mu Emulator X bank and audited, and both output layouts loaded in DirectWave itself - the folder layout and now the monolithic file.

douglas-carmichael:
Simplified: the destination now always writes a monolithic program and the option is gone.

The folder layout had exactly one advantage - the samples stay editable as WAV files - but that is what the WAV and SFZ destinations are for. The monolithic file is a single self-contained file (the point of #70 and of the FL Studio Mobile use case), it is smaller because the embedded audio is FLAC compressed, and it keeps up to 24 bit where the folder layout was capped at 16. Writing one layout also removes the second code path.

Re-verified after the change: 509 programs converted from an E-mu Emulator X bank, 13403 zones, no structural problems and no stray files; and a written program loads in DirectWave with all zones playing correctly.

douglas-carmichael:
Note on the legacy programs mentioned above: their samples are WAV files which contain an Ogg stream instead of PCM, which is why their audio could not be converted. That is fixed separately in #303 (it is a general sample-loading fix, not specific to DirectWave). With both, all 26 legacy DirectWave programs which ship with FL Studio convert completely - 1048 samples, no errors.

douglas-carmichael:
Two defects turned up while converting a real Logic EXS24 instrument (63 zones, three velocity layers) and are fixed in a022813c.

**Written programs could crash DirectWave while loading.** The frame count of every one of the 2040 embedded samples of the 85 monolithic programs I have is a multiple of 512, and the FLAC stream always holds exactly 512 frames more than the frame count in the audio format block. Samples of non-monolithic programs have arbitrary lengths, so the rule belongs to the embedded-audio path. A program whose samples do not fill their last block crashes the plug-in with an access violation a few bytes past the end of a heap block (`EXC_BAD_ACCESS` / `KERN_INVALID_ADDRESS`, 49 bytes past a `MALLOC_LARGE` region in the macOS crash report). The lengths a conversion produces are whatever the source has: resampling 48 kHz material to 44.1 kHz gave 25 ragged lengths out of 63 and crashed FL Studio, while the same instrument converted without resampling happened to land on 128 frame boundaries and loaded. Written audio is now padded with silence up to a full block and the encoded stream gets one block more, which reproduces what DirectWave itself writes.

**Written programs lost the attack of the source instrument.** The creator built its programs from the sample container of an FL Studio Mobile factory program and zeroed the four blocks 0x0069 to 0x006C. Those blocks are not metadata - all 117 programs I have, across all three format versions, carry the same non-zero payloads - and the Mobile container also brings the Mobile defaults of the modulation matrix (Zone LFO 2 to Voice Pitch instead of Pitch Bend to Voice Pitch), of the two LFOs and of the disabled filter. The programs loaded and played, they just did not sound like the source. The preamble and the container template now come from a real monolithic desktop program, so a written file equals a real program byte for byte except for the six blocks which carry the converted content, and the program parameter blocks are written with the payloads the real files have.

Both verified in FL Studio 2026: the instrument which crashed now loads and plays with its attack intact. Round-tripping a real monolithic program reproduces its zone mapping and amplitude envelope exactly, where it previously came out with the Mobile defaults. The design document has the block rule, the program parameter payloads and the decoded modulation matrix indices (the target is the index in the plug-in's name table plus one, which is what makes the desktop defaults read as Note Velocity to Voice Gain and Pitch Bend to Voice Pitch).

PR #313, opened 2026-08-03
Casio FZ: read the head-less FZF full dumps of the fzdump utility (fixes the factory library)

douglas-carmichael:
The FZF files on the Casio FZ archive site (https://zine.r-massive.com/casio-fz-sampler-archive/) - the factory library as well as the Soundwaves, Shareware and Livewire libraries - are written by the 'fzdump' MIDI utility and have **no file head sector**: the content starts directly with the first bank block. The reader stripped the first sector as the head and read garbage counters from it, so every file of the factory library failed with 'Malformed FZ file'.

### How the head-less layout is read

* The FZ keeps its memory bookkeeping in the system parameter area at the end of the first bank block: number of voices at offset 1008, content length in sectors at 1010 and wave blocks at 1012. The bank block count follows as content sectors - wave blocks - voice blocks. A file head is synthesized from these counters, everything downstream is unchanged.
* The two layouts are distinguished by the first block pointer pair: a file head starts with a valid sector range (2 <= start <= end < 1280) or zeroed pointers with plausible counters, while a head-less dump starts with the area count and the ascending high keys of the first bank, which cannot form such a range.
* One factory file (Tenor-Alto Sax of disk FL-6) is truncated: it has 921 sectors but its counters claim 1023. Such files are no longer rejected; the missing wave data is read as silence and a warning names the file and the missing byte count.

### Verification

* All 36 FZF files of the factory library now convert: 153 banks, 1299 samples, no errors. The key/velocity splits, root keys, area volumes, envelopes, filters and the pitch bend range match the values decoded by hand from the hex dumps.
* The same dump with a synthesized file head converts byte-identically to its head-less twin, so the documented layout still reads as before.
* Round trip: factory FZF -> written IMG disk image -> read back detects all banks.

The design notes in CASIO_FZ_FORMAT.md document the layout, the counter offsets and a related observation from the factory voices (the sounding mode is a bit field: 0x0157 = normal without bit 7 occurs and plays normal).


douglas-carmichael:
Extended the verification to every other FZ library circulating on the archive site (Soundwaves, Shareware, Livewire and the two Swedish Users Club CDs, ~9,700 more files) and pushed a second commit with what that surfaced:

* **FZV and FZB dumps are head-less too** and use the same counter scheme: the three counters always sit in the padding at the end of the first 1,024-byte block, no matter whether a bank block, a voice block or (in full dumps without any banks - these exist, 20 files) plain voice blocks occupy it. The head-less recognition now covers all three types, each validated against its expected structure. The Swedish Users Club CDs alone contain 8,840 FZV and 652 FZB files, all of which read now.
* **Bit 15 of the area voice pointer marks an unassigned area** (volume 0, generator count 255, low bits pointing beyond the stored voices - e.g. the literal 'YOUR TURN' placeholder banks of the Soundwaves library). Such areas are skipped silently instead of logging a voice-index-out-of-bounds error.
* **The pitch bend range is only read from full dumps**: every surveyed FZF holds a plausible value there (<= 12 semitones, mostly the default 3), while ~65% of the bare voice/bank dumps carry stale memory in the effect area (values like 179 or 255, which turned into absurd bend ranges). Values beyond 12 semitones are ignored as well.

Final tally across all six corpora: **10,686 multi-samples read with zero errors**; 8 genuinely truncated files degrade to silence with a warning, and one empty 1-sector voice dump reports 'no voices or banks' as it should.

PR #318, opened 2026-08-04
Teenage Engineering OP-XY: fix the loop release mode and the cross-fade scaling, convert the velocity sensitivity (#317)

### LINKED ISSUE #317: Loop Issues With OP-XY implementation
Currently the new OP-XY conversion implementation is still relatively untested and has some issues when encoding to the OP-XY format. The following screenshot comparing 3 conversions of nearly identical Ableton Sampler files will be referenced throughout this issue report. All source Ableton files and resulting OP-XY files are attached in a zip at the bottom of this report.

Image

**Issue 1: Loop Type assignment**
Currently, CwM is assigning the wrong loop type to the resulting OP-XY preset. Shown in the screenshot for "complexOutput1", the "release mode"/loop type set in Ableton is "⇥" which uses the loop region for the sustain, and the plays the region after the loop on release. In the resulting OP-XY preset, the "∞" loop mode is incorrectly assigned which ignores the release region and uses the loop region on release. 

Shown in the screenshot for "complexOutput2", the "OFF" release mode is assigned in the Ableton Sampler. This the exact equivalent of the OP-XY's "infinite loop" mode. Instead the OP-XY's equivalent of Ableton's "⇥" is incorrectly assigned indicated by the lack of the "∞" on the loop setting.

**This means the loop type assignment needs to be flipped when encoding the OP-XY preset file.** Less likely, there is an issue with how Ableton .adg files are interpreted.

important note: it's worth noting that there is a third loop type just called "loop off". This is assigned for one shots. I tested it and this is correctly assigned by CwM when converting.

**Issue 2: Loop Crossfade Assignment**
Next, the loop region assigned to the resulting OP-XY file doesnt smatch the source Ableton Sampler. This likely has to do with how number are arbitrarily scaled within the OP-XY .json preset. You can see this most clearly in the screenshot for "complexOutput3". Here, the Ableton loop crossfade is set to 100% while the resulting OP-XY file only crossfades 17%

some important notes on how this crossfade parameter works: on physical hardware, you can only manually set the cross fade region between 0% and 75%. Thankfully though, you can set any value between 0% and 100% by manually editing/encoding a preset .json file. So the possible range of the loop region parameter needs to be identified from 0% to 100% and then mapped and scaled correctly. Let me know if I can help with this somehow.

important note: While the UI used to show the crossfaded region on the OP-XY visually looks like it would only crossfade data inside the loop region, in reality it crossfades data from the end of the loop region with data just before the loop region, identical to how crossfading works in Ableton's Sampler and most other multisampler players that support crossfading.

here is one more small suggestions when encoding the OP-XY presets:
It would be nice if the preset velocity parameter could be auto-filled from the source preset file or manually set by the user. Something like 64 (as shown on the OP-XY) would be a good number. This corresponds to the "velocity.sensitivity" setting in the .json with 64 shown on the OP-XY corresponding to integer 21299 inside the .json

`"velocity.sensitivity":21299`

**Source and output files for analysis:**
[complexOutput.zip](https://github.com/user-attachments/files/30709247/complexOutput.zip)

douglas-carmichael:
Fixes #317 - all three findings, verified against the attached complexOutput files.

**Loop release mode was inverted.** `loop.onrelease: true` is the '∞' mode of the device: the loop keeps running after note-off. The code treated the flag as a sustain loop ('loop until release'), which is exactly the opposite, so Ableton's '⇥' became '∞' and Ableton's 'OFF' became '⇥'. Creator and detector now invert the flag, and the tag documentation states the actual meaning. Re-converting the files from the issue now yields `loop.onrelease: false` for complexOutput1 ('⇥') and `true` for complexOutput2 ('OFF').

**The cross-fade percentage of the device relates to the whole sample, not to the loop.** The direct evidence is in the issue itself: the value 29368 which was written for complexOutput3 (a 166606-frame sample) is displayed as 17% by the device - exactly 29368 / 166606. A commercial pack of presets in this format corroborates the law (the pack is shipped for the Elektron Tonverk, whose multi-sampled instruments use the same patch.json format - the files declare platform "OP-XY"): every one of its 17 cross-faded regions stores exactly 'display percent × framecount' (73%, 49%, 75%, 32%, 20%, ... - all exact integer percentages of the frame count, the largest being exactly the 75% manual limit of the OP-XY hardware). The stored values regularly exceed both the loop length and the audio available before the loop start, so they cannot be a plain frame count of the fade. The creator therefore writes `crossfade fraction × framecount` and the detector reads the fraction back symmetrically. For complexOutput3 (100% cross-fade in Ableton, which stores 29368 frames of its 30163-frame loop) the written value changes from 29368 (displayed as 17%) to 162215, which displays as ~97% - matching the source.

**The velocity sensitivity of the engine is now converted.** Written from the amplitude-velocity modulation depth of the source when it states one (the Ableton files carry `VolumeVelScale` 0.35 and now produce `velocity.sensitivity: 11468`), and read back by the detector. When the source states nothing, the suggested default 21299 (displayed as 64) is written - full model depth maps to this default as well, since the velocity response of the device at 32767 is much steeper than the neutral 100% velocity tracking of e.g. SFZ.

Verified with a build: the three Ableton files from the issue convert to the expected values above, an OP-XY to OP-XY round trip of the results is stable in all loop fields and the sensitivity, and a plain WAV source receives the 21299 default.

@johnnywagner: the loop-type and cross-fade values should now land as you described. One thing only the hardware can answer: with the new scaling, does a converted 100% Ableton cross-fade also *sound* like the Ableton original (fade over roughly the whole loop), or does the device fade a shorter region? That would tell whether the engine maps the percentage onto the loop or onto absolute frames, and I will gladly adjust the law if you hear a difference.

### DIFF
--- a/documentation/CHANGELOG.md
+++ b/documentation/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Many thanks to Douglas Carmichael for plenty of contributions and fixes!
 * New: Added support for the FL Studio DirectWave format (reading of DWP programs with their key/velocity ranges, gain, panning, loops, amplitude envelope, filter and the pitch and volume LFOs, including monolithic programs which carry all their samples inside of the file; the mapping of DWB banks and of sampled plug-ins is reconstructed from the names of their sample files; written programs are always monolithic, i.e. one self-contained file per instrument which stores all of its samples as FLAC compressed audio).
-* New: Added support for the Teenage Engineering OP-XY multi-sample preset format (reading and writing of the *.preset folders with their patch.json description file).
+* New: Added support for the Teenage Engineering OP-XY multi-sample preset format (reading and writing of the *.preset folders with their patch.json description file, including the loop with its release behavior and cross-fade and the velocity sensitivity of the engine).
 * New: Added support for the Casio FZ-1/FZ-10M/FZ-20M format (reading of floppy disk images (IMG, HFE) and bare dump files (FZF, FZV, FZB - including the head-less layout of the 'fzdump' utility, in which the circulating libraries like the factory library come) with banks, voices, loops, envelopes and filters; writing creates a ready-to-use floppy disk image with a full dump, e.g. for a Gotek/HxC floppy emulator - not yet verified on real hardware).
 * User Interface
   * New: The folder/file history does now remember the selected source format for the folder/file and restores it on selection.
--- a/documentation/README-FORMATS.md
+++ b/documentation/README-FORMATS.md
@@ -902,9 +902,9 @@ file and stores all samples in a sub-folder by the same name. The samples of the
 
 The OP-XY is a portable sampler and sequencer from Teenage Engineering. Its multi-sample presets are open: a preset is a folder with the ending *.preset* which contains all samples as WAV files and a *patch.json* file with the mapping and the settings of the synthesizer engine.
 
-Reading a preset gives the key range, root key, tuning, gain, play range and loop of every region, plus the amplitude envelope and the pitch bend range which the device stores once for the whole preset.
+Reading a preset gives the key range, root key, tuning, gain, play range and loop (with its cross-fade and release behavior) of every region, plus the amplitude envelope, the velocity sensitivity and the pitch bend range which the device stores once for the whole preset.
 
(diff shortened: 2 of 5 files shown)


johnnywagner:
Thanks for getting to this so quickly. If its not too much trouble, could you prebuild another Mac OS ARM version with these changes that I can test? I don't have any coding environments installed on my PC so I think it'd be a pretty steep uphill battle to learn how to do it myself

PR #356, opened 2026-08-19
Waldorf QPAT: Apply the de-click minimum to an amplitude release which the source left unset

douglas-carmichael:
#166 added a minimum amplitude attack/release so the device no longer clicks on an instant stage:

```java
private static double declickAmpTime (final boolean isAmplitude, final double seconds)
{
    return isAmplitude && seconds > 0 ? Math.max (seconds, 0.07) : seconds;
}
```

The `seconds > 0` guard means a release of exactly zero - the hardest-clicking case, and the one the minimum exists for - is written through unchanged. A source which leaves the release unset lands there through the clamp above the call:

```java
final double releaseTime = declickAmpTime (isAmplitude, Math.clamp (envelope.getReleaseTime (), 0, 60));
```

`getReleaseTime ()` returns -1 when unset, `Math.clamp` makes that 0, and the guard then skips the minimum. A zero release gates the VCA off in one sample and cuts the waveform wherever it stands - for a looped zone that is always full level, so the preset clicks on every key release.

Reproduced with the Ensoniq ASR-10 'ASR ORGAN' patch from the commercial Digital Sound Factory EXS24 library, whose envelope sets a decay but no release:

| | attack | decay | release |
|---|---|---|---|
| before | 0.00 secs | 10.00 secs | **0.00 secs** |
| after | 0.00 secs | 10.00 secs | **0.07 secs** |

Confirmed audible on an Iridium MK2 - the preset clicks on release before the change.

The attack keeps the existing behaviour and still passes a zero through, so percussive material keeps its instant onset: an attack starts from silence at the beginning of the sample, while a release always has a sounding waveform to fade out. The change is therefore a separate method rather than a relaxed guard, and `0.07` becomes a named constant since both use it.

Not rare in practice: across four commercial libraries converted here, 42 presets write a zero amplitude release, and all but a handful have looped zones.


douglas-carmichael:
Hardware result from an Iridium MK2, which also corrects one of my assumptions.

Confirmed: a preset written with a zero amplitude release clicks on every key release, and raising the release removes it. The Ensoniq 'ASR ORGAN' patch was re-imported unchanged and still clicked; setting the release to 0.18 s on the device silenced it, with nothing else altered.

That means the device does not treat parameter value 0 as the 0.06 s minimum the conversion assumes:

```java
// creator
if (y <= 0.06) return 0;   // "The minimum representable time is 0.06 seconds (parameter value 0)"
// detector
return 0.06 * Math.pow (1000, x);   // x = 0 gives 0.06
```

If 0 really were 60 ms the preset could not click the way it does. The device appears to gate instantly at 0, so writing 0 is what has to be avoided - which is what this change does.

What I have not yet established is whether `0.07` is far enough above it. The value verified on hardware is 0.18 s; 0.07 s is untested, and it sits just above the 0.06 s the law calls its minimum. That matters beyond this patch, because #166 already clamps any non-zero amplitude attack/release to the same 0.07 - so if 0.07 still clicks, that clamp is not achieving its purpose either.

I am running a threshold test on the device with presets at 0.07, 0.12 and 0.18 s and will report the smallest value which is reliably clean. Happy to move the constant to whatever that turns out to be.

douglas-carmichael:
Threshold measured on the hardware: **0.07 is enough**, so the constant in this patch stands as written.

I recorded the output of an Iridium MK2 playing the same preset at a 0.07 s release and analysed the note-off two ways. Neither shows a click.

A click is broadband, so high-passing above 4 kHz and following that energy through the release: it falls with the signal (-5, -9, -10 dB) instead of spiking. A click would do the opposite - high frequency jumping while the level collapses.

A click is also a step much larger than the waveform's own motion at that level:

| region | max sample step | RMS level | ratio |
|---|---|---|---|
| sustain, key held | 150.5 | 243.7 | 0.62 |
| release, first half | 17.0 | 175.6 | 0.10 |
| release, final collapse | 14.0 | 70.8 | 0.20 |

The release moves *less* than the note does while it is sounding. The largest step anywhere in the recording is mid-sustain, part of the organ's own chorus, at ratio 0.52.

The same session also confirms the time law itself, which I had wondered about above. Presets written at 0.07, 0.12, 0.18 and 0.50 s read back on the device as 0.069, 0.119, 0.179 and 0.499 s - `0.06 * 1000^value` reproduces all four to within a millisecond, the shortfall being float32 storage and the device truncating rather than rounding. So the conversion is right across its range, for attack and decay as well.

That leaves exactly one anomaly, and it is the one this patch addresses: the law says value 0 should be 0.06 s, but the device gates instantly there. Value 0 is a special "off" case rather than the bottom of the scale, which is why a preset written with a zero release clicks while one written at 0.07 does not. The comment in `convertFromTime` describing 0.06 s as "the minimum representable time (parameter value 0)" is accurate as a limit but not at exactly 0; I have left it alone here since it is only the release write that needs to change.

Also worth noting for #166: since 0.07 clears it audibly, the clamp added there was the right value all along - the only gap was the `seconds > 0` guard letting a true zero through.

PR #364, opened 2026-08-20
Akai S-1000: read the sample rate of the samples recorded at 22050Hz

douglas-carmichael:
An S-1000 sample recorded at 22050Hz was read with a sample rate of **0Hz**, which broke the conversion in two ways.

With the sample rate processing option enabled, the re-sampling divides by that rate, so `44100 / 0.0` becomes infinite, `Math.round` returns `Long.MAX_VALUE` and the cast to `int` truncates it to -1:

```
java.lang.NegativeArraySizeException: -1
    at ...algorithm.SincResampler.resample(SincResampler.java:60)
    at ...algorithm.AudioSampleReducer.resampleFrequency(AudioSampleReducer.java:400)
    at ...core.ConverterBackend.processSamples(ConverterBackend.java:791)
    at ...core.detector.AbstractDetector.handlePresetFile(AbstractDetector.java:376)
```

A disk image delivers all of its presets from one file, so that error ends the loop over them: the conversion of *East West - Twisted Textures CD1* stopped at preset 93 and the remaining **32 of its 124 presets were never written**. The log gives only a bare `-1`, so nothing points at the sample or even at the preset which caused it.

Without the processing option the conversion completes, but the affected samples are written as files which claim a sample rate of 0Hz and are therefore unplayable.

## Cause

The sampler fills in the exact sample rate only for the samples it records at 44100Hz. The samples recorded at 22050Hz leave that field at zero and are marked solely by the index at the start of the sample header - the field the reader skipped:

```java
// 0 for 22050Hz, 1 for 44100Hz - skip
disk.readInt8 ();
```

The index is now used as the fall-back whenever the exact sample rate is missing. Nothing changes for a sample which carries it.

## The two fields never disagree

Across the 4088 sample headers of four commercial CD-ROMs:

| index | exact sample rate | samples |
| --- | --- | --- |
| 1 | 44100 | 4063 |
| 0 | *missing* | 25 |

There is no header which states one rate in the index and a different one in the exact field, and no header where the index is missing. The 25 samples which only carry the index are the ones which were unreadable.

## Verified

*East West - Twisted Textures CD1*, all 149 presets to Waldorf Quantum/Iridium:

| | before | after |
| --- | --- | --- |
| presets written with re-sampling to 44100Hz | 93 of 149, then aborted | **149, no error** |
| `VCTRY24.wav` written unprocessed | 1 ch, **0 Hz**, 247163 frames | 1 ch, **22050 Hz**, 247163 frames (11.21 s) |
| `VCTRY24.wav` re-sampled to 44100Hz | *crash* | 1 ch, 44100 Hz, **494326 frames** (11.21 s) |

The re-sampled frame count is exactly twice the source, which confirms the 22050Hz reading.

*Twisted Textures CD2* (121 presets), *CD3* and *Best Service XX Large - Pads* (335 presets) convert unchanged and without errors - none of them contains a sample which is missing the exact rate.


git-moss:
@douglas-carmichael Could this be the reason for the MV-8000 conversion issue?

PR #367, opened 2026-08-20
Waldorf Quantum/Iridium: keep the instant attack of a percussive sound

douglas-carmichael:
This corrects an over-correction of my own, from #166.

The device cannot play an amplitude stage shorter than **0.07 s** (`y = 0.06 × 1000^value`), so a shorter one collapses to an instant stage, which snaps the VCA open and clicks when the audio does not start at a zero crossing. #166 fixed that by lifting every short amplitude stage to 0.07 s.

Applied unconditionally, that lift is wrong for the normal case. A sampled instrument carries its attack **in the audio** and starts at silence, so there is nothing to click - the lift only fades the recorded attack in over 70 milliseconds and takes the strike off every piano, plucked string and drum.

An Akai S-1000 electric piano asks for **3 ms** and was written as **70 ms**, 23 times longer. It stopped sounding like a piano; that is how this was found.

## An instant attack is normal on this device

The manual: *"If Attack is 0 and Decay set to a medium value you get a percussive sound; if you turn up the attack, you get a soft-sounding start."*

Measured over 141 presets written by an Iridium:

| amplitude attack | presets |
| --- | --- |
| exactly 0 (instant) | 50 (35 %) |
| above 0, below 0.07 s | 35 (25 %) |
| 0.07 s or longer | 56 (40 %) |

60 % use an attack shorter than the lift forced, and a third use exactly the instant stage the lift was written to avoid.

## Change

The audio of the zones is measured, and the stage is lifted **only when a zone really starts with a step which can be heard**: more than 2 % of the peak level of that audio - the ratio at which the wrap of a loop is treated as audible elsewhere in the code (`LoopClickDetector.MINIMUM_RELATIVE_STEP`). Measuring against the peak of the same audio keeps it independent of the bit resolution.

The measurement runs only for a preset whose attack is short enough to be affected at all, so a preset with a genuine attack reads no audio. Unreadable audio keeps the safe, non-clicking variant. The release is unchanged: a note-off lands at an unpredictable point of the audio, so its step cannot be measured in advance.

## Verified

117 presets of an Akai S-1000 CD-ROM converted to Quantum/Iridium:

| | presets |
| --- | --- |
| instant attack restored | **78** |
| genuine longer attack, untouched | 38 |
| lift kept because the audio would click | **1** |

The one that keeps the lift is `TINESNPAD B`, whose worst zone starts at **3.49 %** of its peak (first frame 970). `DIGIJAZZ`, the electric piano, starts at **0.43 %** and now gets its instant attack back.

Whole-disc conversion time is unchanged at ~4.8 s.


git-moss:
@douglas-carmichael Please resolve the conflicts, I am unsure which code is the current one.

douglas-carmichael:
Rebased onto current main - same situation, a textual collision rather than two versions of the same fix.

This branch predates #356, and both change the de-click of the amplitude envelope in `WaldorfQpatCreator`. #356 made the **release** always get the minimum, even when the source leaves it unset; this one makes the **attack** keep its instant rise when the material does not need de-clicking. Merged, both hold:

- release: `declickAmpRelease (isAmplitude, ...)` - unchanged from main, still applied even at zero
- attack: `declickAmpTime (isAmplitude && !allowInstantAttack, ...)` - only lifted when `startsWithAudibleStep` says the audio starts with a step

Current is main, and this adds the audio inspection on top. Verified on two sources with an identical 10 ms attack: one whose audio fades in from silence keeps `AmpEnvAttack = 0.01 secs`, one whose first sample is a large step is lifted to `0.07 secs`. The renaming of the parameter of `declickAmpTime` from `isAmplitude` to `declick` is only because the flag is now more than "this is the amplitude envelope".

PR #368, opened 2026-08-20
Akai S-1000: shape the envelope of the loudness like the sound hardware

douglas-carmichael:
The envelope of the loudness was written as a straight line in loudness. The sampler moves it as a straight line in **decibel**, so a converted preset holds a note near its full level for most of a stage and then drops it. An electric piano stops sounding like one, which is how this was found.

## Read from the v4.4 operating system

The level of a voice is an **index into the gain table** at `0x5E10`, and the operating system adds the rate of the envelope to that index once per tick:

```asm
14C82  mov  cx, word ptr [si + 0xac]   ; the envelope of this voice
14C86  add  cx, ax                     ; += rate, a straight line in the index
14C88  mov  word ptr [si + 0xac], cx
14C91  sub  ah, ah
14C93  call 0x14d13                    ; gain = table[index]
14C96  mul  word ptr [si + 8]          ; times the level of the voice
14CA2  out  dx, ax                     ; to the sound chip
```

The table it lands in is exponential - measured over its own entries, **0.237 dB per index step**, 60.5 dB across its 255 steps:

| index | value | dB below full |
| --- | --- | --- |
| 128 | 1024 | -30.10 |
| 160 | 2452 | -22.52 |
| 192 | 5873 | -14.93 |
| 224 | 14063 | -7.35 |
| 255 | 32767 | 0.00 |

The lookup itself treats the index as decibel too - an attenuation is **subtracted from it** before the table is read, and the variants of the call pass `[si+0xaf]` for a stage which ends above silence and `0xff` for one which ends at it:

```asm
14D1F  mov  al, cl        ; the index
14D21  sub  al, ah        ; minus an attenuation, in the same units
14D37  mov  ax, word ptr [di]
```

A straight line in that index is a straight line in decibel, so the loudness follows a curve.

## What it changes

'DIGIJAZZ', an electric piano with a decay of 9.5 seconds to silence:

| level | written before | the sampler |
| --- | --- | --- |
| -6 dB | 4.74 s | **0.94 s** |
| -12 dB | 7.12 s | **1.89 s** |
| -20 dB | 8.56 s | **3.14 s** |
| -40 dB | 9.41 s | **6.29 s** |

## The filter envelope is deliberately left alone

Its index addresses the cut-off table, which is exponential in frequency, so a straight line there already is a straight line in the cut-off parameter of a destination. Only the envelope of the loudness is shaped; the amplitude and the filter envelope share one conversion, so the shape is applied to the amplitude one after it is built.

Written to a Waldorf Quantum/Iridium the amplitude curves become `Exp` and the filter curves stay `Lin`. The 149 presets of a commercial CD-ROM convert to SFZ and to DecentSampler without an error as well; both formats carry the slope, so they gain the correct shape too.


git-moss:
@douglas-carmichael Please resolve the conflicts, I am unsure which code is the current one.

douglas-carmichael:
Rebased onto current main - the conflict was textual, not a disagreement about the code.

The branch was written before #359 went in, and both touch the same part of `AkaiS1000ProgramConverter`: #359 added `scaledLoudnessToDecibel` and `scaledSustainToLevel` right where this one adds `applyLoudnessCurve`, so git showed the two new methods as competing versions of one block.

They are independent and both are wanted. #359 fixes **how long and how loud** a stage is (the rate table, the loudness scale, the sustain as an attenuation); this one fixes **the shape within a stage** - the hardware adds the rate to an index into an exponential gain table, so a stage is a straight line in decibel, and writing it as a straight line in loudness keeps a note near full level for most of the stage and then drops it.

Current is main. This branch is now main plus 26 added lines and nothing removed: the `applyLoudnessCurve (amplitudeEnvelope)` call and the method itself. Builds clean, and re-analysing the two Akai ISOs reads all 1245 programs without error.

PR #393, opened 2026-08-26
E-mu Emulator II: read bank files (EII), read the loops as the sampler plays them, cut voices at the bank memory the sampler saves

douglas-carmichael:
Your test of the factory library brought up two issues - the `.eii` files of 485,887 bytes were rejected for their size, and 24 disks reported a bank which continues on another disk - and settling the second one exposed a third, which is the largest: nearly every looped voice lost its loop.

**Bank files.** `.eii` is the *Sound Designer for EII* file: the bank memory alone, from the bank address 0x9600 to the end of the 512 KB address space, less one byte (0x80000 − 0x9600 − 1 = 485,887). EMXP's CDS3TOOL extracts them from the CDS3 CD-ROMs, EMXP writes them, and Arturia's Emulator II V imports them, which is why libraries circulate in this form. They are read now. The bank is located from the self-pointers of the voice records (their high byte is the page of the record, 0x9B for the first), so a file with a header in front of the bank is read as well.

**The end of the bank.** The sampler saves exactly those 485,888 bytes: the last 8,704 bytes of the bank region of every one of the 90 factory disks are zero, also on the disks whose last voice reaches the end - and they were read as audio of that voice. A voice which was sampled into the memory behind that limit - 124 voices on 32 factory disks, from 121 frames (Grand Piano) to 26,859 (Orchestra Tune) - is cut there and noted; the rest of its audio is on no disk. None of the messages of OS 3.1 knows a bank on two disks (its "Insert Another Disk" prompts belong to the disk copy and format functions), so the note no longer claims one.

**Loops.** Checking the voice fields against the playback counters which the sampler stores next to them (start − 1, negative length, end − 1, negative loop length) showed the reading wrong: the loop switch is bit 2 of the flags, not bit 1, and the field which was taken for the sample end is the loop start. On the Emulator II the loop is the end of a voice - the audio in front of it plays once, the loop repeats - which the audio confirms (the region behind the "sample end" of `Piano A 2` continues the decay of the note at its pitch). 3,311 of the 3,848 factory voices are looped; before, 274 of them looped their beginning and all others ended at their loop start. The counters agree with the new reading on 3,727 of the 3,768 voices whose counters are valid, the other 41 differ by 4 frames; the 80 voices whose counters are stale (flag bit 4) are read from the fields like all others. Voices which start their playback later than their memory (the sustain-only copies of the stacked strings) and voices which are windows into one shared recording (Mallet Cymbals, where the loop fields describe the whole recording) are read as the sampler plays them. The creator writes the same layout - the audio up to the end of the loop - and no longer flags unlooped voices as looped (its default flags were 0x24, which is looped).

Results over the 90 factory disks + OS disk: 605 presets, 5,214 zones, 4,254 of them looped, no errors. The `.eii` files for the test were carved out of the factory images; they decode to the same zones as the disks, less the one frame their missing last byte held. Grand Piano converted to SFZ and back to a disk reads back with identical lengths, loops, key ranges and roots. Written disks remain untested on hardware. The design document has the corrected record layout.


douglas-carmichael:
Second commit, from testing against real `.eii` files: archive.org's "Emu II Samples Arturia" holds 1,481 Sound Designer bank files of the community library (OMI Universe of Sounds etc.), all 485,887 bytes; 1,325 of them are byte-identical to the bank region of their namesake disk image in the HxC corpus, which settles the layout of the file.

That corpus exposed one more reader bug, independent of the bank files: a preset record was recognised by an ASCII name, and the chain stopped at the first record which failed it - the signature presets of the Northstar/OMI libraries ("p&c© 1986 N*", Mac Roman from the Macintosh software) and names with a stray tab or carriage return. 142 of the 1,481 banks lost 792 presets that way, and the factory disk *DAS Synth* 12 of its 13 (a name ending with a carriage return). A record is now recognised by its structure (header bytes, key ranges with counts covering at most the 61 keys), which every real record has and the leftovers behind a chain never have; names are decoded as Mac Roman. Results: factory disks 617 presets (was 605), the 1,481 bank files 7,720 (was 6,928), no bank loses a preset or a zone. The note about the bank memory now counts only voices which a preset plays.

git-moss:
@douglas-carmichael I already merged the first commit, so I think this one did not get into it.

douglas-carmichael:
Third commit: the voice settings are decoded - filter (cutoff, Q, keyboard tracking, envelope, velocity amount), VCA and VCF envelopes, LFO (rate, delay, depths on pitch/filter/level), velocity to level and to attack, fine tuning and level. The record keeps them the way the playback engine uses them (velocity tables, envelope counters on a 10 ms tick, pointers into depth tables); the bytes and laws come from pairing 9,563 voice records of the factory and OMI libraries with the zones of EMXP's SoundFont conversions of the same banks. Cross-check through the model: every preset of the 90 factory disks read from the disk and from the SoundFont agrees on 96-99 % of every attribute (3,543 zones). The design document has the byte map and the laws. Chorus and loop-in-release are still not decoded.

The SoundFont side of that comparison needs #394 (coarse offsets, global-zone modulators).

PR #405, opened 2026-09-09
Audiomodern Soundbox: the band-pass and the high-pass of a filter effect were swapped, which silenced converted presets

douglas-carmichael:
Converting a preset of the Audiomodern *Wired* pack to a Waldorf Iridium produced a patch which plays no sound at all. `WRD PAD - Canadi pad` carries a high-pass at 61.7 Hz - a plain rumble filter - and it arrived on the device as a **12 dB band-pass at 61.7 Hz**, roughly 34 dB down at 440 Hz and 60 dB down at 2 kHz, so the whole pad is gone.

`SoundboxFilterEffect` mapped the `TYPE` attribute of the filter effect onto the ordinals of `FilterType` (low-pass, high-pass, band-pass, band-rejection). That order was an explicit assumption in the format description, because every pack of the original corpus (*Cosmo*, *Kromium*, *Voxmotive*, *Starlit*, *Mechania*) uses type 0 only, so nothing contradicted it.

Soundbox uses JUCE's order instead. The plug-in binary (1.2.x) holds the parameter as `AMParameterEnum<juce::dsp::StateVariableTPTFilterType>` - the mangled symbol `15AMParameterEnumIN4juce3dsp26StateVariableTPTFilterTypeEE` is in `Soundbox.vst3` - and JUCE declares

```cpp
enum class StateVariableTPTFilterType { lowpass, bandpass, highpass };
```

so **0 = low-pass, 1 = band-pass, 2 = high-pass**, and there is no band-rejection at all.

The cutoffs of the *Wired* pack agree: its 44 type 0 slots have a median cutoff of 5.4 kHz (low-pass), the 23 type 2 slots a median of 106 Hz with a minimum of 50 Hz (high-pass, i.e. rumble filters) and the 10 type 1 slots a median of 543 Hz (band-pass). One of the type 2 presets is named `WRD SYNTH - BrokenHP`.

Both directions are fixed: reading maps 0/1/2 onto the three types the plug-in has, writing maps them back. A band-rejection of the model is written as the low-pass default, since the plug-in cannot store one - it previously wrote a `TYPE` of 3, which is outside the enumeration.

**Effect on the pack:** 20 of the 40 *Wired* presets carry a type 2 filter. Re-analysing the whole pack before and after gives 319 band-pass and 319 high-pass zones exactly swapped, 605 low-pass unchanged, and no other difference anywhere in the analysis. The 13 presets whose cutoff sits between 50 and 130 Hz were silent; the ones at 737 Hz, 1460 Hz and 1940 Hz were thin and nasal.

A Soundbox to Soundbox round trip of `Canadi pad` now writes back `FREQUENCY="61.69" RESONANCE="2.2730" TYPE="2.0"`, matching the source; before the fix it turned the high-pass into a band-pass. The silence was found on an Iridium MK2; the corrected patch reaches the device as `Filter12Type` "12dB HP" at 61.7 Hz, but has not been played on the device yet.

The format description in `documentation/design/SOUNDBOX_FORMAT.md` is updated with the enumeration and the evidence.


douglas-carmichael:
Correction to the opening paragraph: the band-pass is **not** what silenced `WRD PAD - Canadi pad`. With this fix applied the preset was still silent on the device.

The silence comes from a second, independent bug, now #407: the sample positions of a pack whose samples are FLAC were all lost, so every loop became a loop of zero length at the start of the sample and the device never advances through the audio. The *Wired* pack holds FLAC.

Everything else in this PR stands - the filter type mapping really is JUCE's `lowpass, bandpass, highpass` and the evidence for it is unchanged, a 61.7 Hz high-pass really was being written as a 61.7 Hz band-pass, and the write path really did emit an out-of-range `TYPE` of 3 for a band-rejection. Only the claim that this alone accounted for the silence was wrong; on its own it would have left the pad heavily attenuated rather than silent.

The two fixes are independent and do not touch the same code.

PR #452, opened 2026-09-19
Roland ZEN-Core: read the WAV files which sample packs store scrambled

douglas-carmichael:
The `02_Melodic.svz` you were sent (a samples-only pack of 32 stereo loops, model tag `Nemly`) converts to full-scale noise. It is the file which #204 was built from, and #204 misread it.

Its SMPd chunks are not raw PCM behind a 158-byte header. Like the files of Roland's SVZ Sample Converter they hold the 0x20-byte header followed by a complete WAV file - `fmt ` (18 bytes), `fact`, `data` and a trailing `smpl` chunk - but every byte of the WAV file is substituted. A stored byte decodes as its bits re-ordered and multiplied by 0xDD modulo 256, with one bit order for the bytes at even and one for the bytes at odd offsets from the start of the WAV file, i.e. the low and the high bytes of the samples. The 158 "header" bytes were all the bytes of the chunk which are not audio (32 header + 58 WAV header + 68 `smpl`), so the reader took the scrambled bytes, shifted by 17 frames, as PCM.

The reader now tries, in this order: a plain embedded WAV (the converter's files, as before), a scrambled one (decoded, then read with the same WAV reader), and the FANTOM's raw PCM behind the 460-byte header. The compact-header code, which only ever matched this file, is removed. Writing is unchanged.

Verification:

- All 32 samples decode to valid WAV files whose `data` sizes and `smpl` loop ends match the `USPa` records of the SVZ.
- The audio is music: the RMS of the frame-to-frame difference is 0.04-0.34 of the signal RMS, where white noise gives 1.41 and the current output 1.10-1.36 (at an RMS of about 18,900 and a left/right correlation of 0.00). Every loop peaks at the same 31129, the level the pack was normalized to.
- The converted samples are identical to the output of a decoder written independently of the application.
- The 188 other SVZ files at hand (device exports, converter output, ZENOLOGY and ConvertWithMoss files) give 4,102 WAV files which are byte-identical to an unmodified build of main.

How the law was found is in ZENCORE_FORMAT.md §3.4, together with what is not known: neither Roland's SVZ Sample Converter 1.00 (Windows and macOS, it has no scrambling code) nor the FANTOM-0 1.08 firmware contains a table of it, so this one file cannot tell whether other packs use other laws. Such a pack is now reported as having no user samples instead of being converted to noise. The document also notes that the two "version" bytes of the file header are block counts - the converter computes them that way - which is why this file carries `02 01`.


git-moss:
Thanks!

